PR #70, opened 2015-02-27
Add link to gist for ES7 class properties proposal

jeffmo:
Adds a link to the [in-progress] gist I'm working on for the ES7 Class Properties proposal


rwaldron:
Thanks @jeffmo !

PR #76, opened 2015-05-06
Added SIMD.js to agenda for May-2015 meeting

PeterJensen:


PeterJensen:
@allenwb Is it OK to add his agenda item?

PR #80, opened 2015-05-23
Update 05.md

littledan:
Link to proposed SIMD.js draft spec

### DIFF
--- a/2015/05.md
+++ b/2015/05.md
@@ -34,7 +34,7 @@
   1. ECMA-402 2nd Edition
   1. ECMA-262 7th Edition and beyond
     1. Shared-memory TypedArrays, atomic operations on shared memory cells, and atomic operations for wait and wakeup.  [spec. proposal](https://docs.google.com/document/d/1NDGA_gZJ7M7w1Bh8S0AoDyEqwDdRh4uSoTPSNn77PFk/edit?usp=sharing) (Lars Hansen - Mozilla)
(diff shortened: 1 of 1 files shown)


rwaldron:
@littledan where is the draft spec source? Specifically: if I need to file issues or pull requests to propose changes, what repo contains the document that I would make those changes or fixes in?

ljharb:
Looks like https://github.com/littledan/littledan.github.io/blob/master/simd.html

rwaldron:
That's what I thought as well, but then came across this: https://github.com/littledan/value-spec/blob/master/spec.html

littledan:
Spec.html is the source, simd.html is the generated output.

rwaldron:
Thanks.

rwaldron:
@littledan I just noticed a link in that source, directing issues to be filed in yet another repo.

littledan:
You're right, it's confusing to have these all in different repos. I actually have a copy of everything in John's repo. Let's say that the authoritative one is there, with the source at https://github.com/johnmccutchan/ecmascript_simd/blob/master/tc39/spec.html . My repo will just hold my own drafts.

littledan:
On the other hand, the web link can still point to the same location, and I'll just make sure that the website is kept in sync with the authoritative copy in John's repo (since I don't know of any other ecmacscript_simd website for me to put them at).

PR #84, opened 2015-05-27
Update 05.md

leebyron:
Since our last TC39 meeting, the spec text for export-from additions have been fully specified, ample discussion has happened with module proposal champions, and babel has included transpilers for them. I'd like to get this on the agenda to discuss advancing from stage 1 to stage 2 or 3.


leebyron:
Thanks Rick!

PR #92, opened 2015-07-08
Update 07.md

ecmageneva:
I try to introduce my points into the agenda.
No idea how to make 3.1 instead of 4...

### DIFF
--- a/2015/07.md
+++ b/2015/07.md
@@ -24,6 +24,8 @@
     1. Host facilities, local logistics
   1. Adoption of the agenda (TODO: Ref document name)
   1. Approval of the minutes from May 2015 (TODO: Ref document name)
+  2. Report from the Ecma Secretariat
+  3. Status of ISO/IEC Fast Track of ECMA-262 Ed.6, ECMA-402 Ed.2 and ECMA-404
   1. ECMA-262 7th Edition, 2016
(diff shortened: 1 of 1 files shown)


bterlson:
Sorry missed this pr. Will add these items tomorrow since this won't apply cleanly anymore.

ecmageneva:
Thanks...
But how could I have made the edits myself?
This looks very complicated.

bterlson:
Easiest way to do it would have been to just make the edits again starting from the most recent version and ditch this PR (which is what I'm doing for you now). There are fancier ways...

Also, numbered lists in this format (called Markdown) start with any digit and a dot and are numbered automatically. You make sublists by indenting. Example:

```
1. first
    1. first first
    1. second first
1. second
```

1. first
    1. first first
    1. second first
1. second

bterlson:
Ok the latest agenda should contain your items. Please take a look.

ecmageneva:
Brian,
many thanks that already helps a lot.
But I would also like to play a bit with Github in order that I get a better feeling what it is and how it works. Where I have my next difficulties is that I have installed Github for Windows, but I do not find how I could make the local copy of the agenda on which I can make the changes, and after how to feed it back to Github.
Kind regards,
Istvan

bterlson:
I think it's best to not take a local copy for just editing the agenda. You can do it entirely from the website. Here are the instructions: https://help.github.com/articles/editing-files-in-another-user-s-repository/.

PR #96, opened 2015-07-14
Update 07.md

littledan:


domenic:
I assume because of web compat or similar? Link?

littledan:
Because of web compat. There was discussion here: https://esdiscuss.org/topic/number-prototype-not-being-an-instance-breaks-the-web-too . Mark Miller suggested just rolling this back for Number, and other possibilities were raised in that thread, but I'm wondering if we should just roll back the semantics of all prototypes to ES5. I don't really have anything more than that to write up.

domenic:
Got it. @bterlson does Edge do Number.prototype as an ordinary object? And has it not encountered this problem?

PR #100, opened 2015-07-27
Adding link to Observable spec

jhusain:
...detailing proposed changes to Observable spec. Also removing topic (scarce resource disposal in generators) in apprehension of discussing it during Promise cancellation.


bterlson:
Would be good to put your name next to the agenda items you own.

Also I am ok leaving the scarce resource topic on the agenda - we can skip if we have already covered it. Up to you :)

PR #113, opened 2015-09-21
Add proposals: String#{padLeft,padRight}, Object.{values, entries}, String#matchAll

ljharb:
I also tweaked the markdown numbering for the rest of the items so that if there's future reordering, it won't create as much diff churn.


michaelficarra:
> I also tweaked the markdown numbering for the rest of the items

Thank you! This had been bugging me.

PR #115, opened 2015-09-23
Discuss topics with written proposals first?

littledan:
@wycats how would you feel about this?


bterlson:
Is this still needed @littledan

littledan:
No--I discussed this with @wycats offline and we ended up doing it.

PR #123, opened 2015-11-08
Add agenda item for Symbol.species improvements.

zenparsing:
There are some consistency issues with how Symbol.species is used across the different built-ins, and there are some questions about whether new built-in constructors that support subclassing should have a species method.  Also, there is a suggestions that it might be better just to put a default Symbol.species implementation on Function.prototype.


domenic:
Mind rebasing?

zenparsing:
Good now?

PR #125, opened 2015-11-13
Add other reported issues

rossberg-chromium:


littledan:
@bterlson What do you think of these agenda items?

bterlson:
Bah sorry I missed this. We are going to take this up, don't worry.

PR #156, opened 2016-01-28
strikeout class fields agenda item

jeffmo:


ljharb:
you may want to delete it - `~~` usually means it's been discussed

jeffmo:
ok, updated

flying-sheep:
so it wasn’t? :disappointed: 

what are we going to do now regarding jeffmo/es-class-fields-and-static-properties#26?

michaelficarra:
January meeting is over. Let's close this.

flying-sheep:
for people ending up here and wondering what’s going on:

automatic semicolon insertion (ASI) applies for the class fields as normal, so no semicolons required.

PR #157, opened 2016-01-29
Fix markdown

chicoxyzzy:
close strikes


chicoxyzzy:
Should I do something else here?

rwaldron:
Nope, that's great—thank you!

PR #175, opened 2016-03-31
Add public class fields to agenda

jeffmo:
Would like to give an update on the status of public class fields if there's time


ljharb:
LGTM

PR #178, opened 2016-03-31
Spell the name of Microsoft's home town correctly.

FUDCo:

PR #180, opened 2016-03-31
Spell the name of Microsoft's home town truly correctly

jfbastien:
Because https://github.com/tc39/agendas/pull/178 was almost correct.


ljharb:
Done in #179

PR #190, opened 2016-05-23
Add private fields to the adgenda

zenparsing:


zenparsing:
Whoops, put it in the wrong section.

zenparsing:
Good now.

PR #193, opened 2016-06-01
Add agenda template

bterlson:
Thought I'd add an agenda template with some of the changes discussed during and after the last meeting. This is merely a first stab, feel free to tear it apart.

### DIFF
--- a/template.md
+++ b/template.md
@@ -1,0 +1,51 @@
+<img src="../images/Ecma_RVB-003.jpg" align="right" height="70" alt="" />
+
+## Agenda for the TBD meeting of Ecma TC39
+
+- **Host**: TBD
+- **Dates**: TBD
+- **Times**:
+  - 10:00 to 17:00 PDT on TBD
+  - 10:00 to 16:00 PDT on TBD
+- **Location**: TBD
+- **Wifi**: TBD
+- **Dinner**: TBD
+- **Contact**:
+  - Name: TBD
+  - Phone: TBD
+  - Email: TBD
+
+## Logistics
+
+TBD Host Logistics (Parking, building access, etc.)
+
+### Registration
+
+TBD
+
+### Hotels
+
+TBD Hotels nearby (optional)
+
+## Agenda items
+
+1. Opening, welcome and roll call
+  1. Opening of the meeting (Mr. Neumann)
+  1. Introduction of attendees
+  1. Host facilities, local logistics
+1. Adoption of the agenda
+1. Approval of the minutes from last meeting
+1. Report from the Ecma Secretariat (15m)
+1. ECMA262 Status Updates (15m)
+1. ECMA402 Status Updates (15m)
+1. Test262 Status Updates (15m)
+1. Timeboxed agenda items: topics are timeboxed in increments of 15 minutes up to 60 minutes. Shorter proposals come first. Additional discussion comes at the end of the meeting.
+1. Stage 0+ proposals looking to advance
+1. New proposals
+1. Discussion and updates for Stage 0+ Proposals
+1. Overflow from timeboxed discussion items (in insertion order)
+1. Closure
+
+## Dates and locations of future meetings
+
+TBD (Copy the table from the previous agenda)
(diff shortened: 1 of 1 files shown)


template.md
```diff
@@ -0,0 +1,51 @@
...
+  1. Opening of the meeting (Mr. Neumann)
+  1. Introduction of attendees
+  1. Host facilities, local logistics
+1. Adoption of the agenda
+1. Approval of the minutes from last meeting
+1. Report from the Ecma Secretariat (15m)
+1. ECMA262 Status Updates (15m)
+1. ECMA402 Status Updates (15m)
+1. Test262 Status Updates (15m)
+1. Timeboxed agenda items: topics are timeboxed in increments of 15 minutes up to 60 minutes. Shorter proposals come first. Additional discussion comes at the end of the meeting.
```
ljharb: shouldn't these also be grouped by "advance, new, updates"?

template.md
```diff
@@ -0,0 +1,51 @@
...
+1. Stage 0+ proposals looking to advance
```
michaelficarra: "stage 0+" is strange. Why not use the term "existing"?

template.md
michaelficarra: Agreed. Also, why limit it to 60 minutes? Until we see how it works, we should enforce as few additional constraints as possible.

ljharb: I'd prefer to throw the TimeboxError for now, and lift the restriction later, rather than try to tighten it up later.

michaelficarra:
Inspired by @msaboff's suggestion in the timeboxing thread, I think we should add spots for

* note-taking volunteers
* technical presentation requirements (connector/adapter, Google hangouts account, etc.)
* a deadline for adding items to the "looking to advance" section

template.md
bterlson: I think it's sufficient to simply order by timebox size since the timeboxed items will be gone through quickly. Since we're pretty much assured to make it through all the timeboxed items, ordering by topic type doesn't seem very useful to me. I don't mind much though, so I'll change it if no one agrees with me!

Regarding the timebox limit, my feeling is that if you can't say that your discussion will be <= 1 hour then you really don't have much idea how long the discussion should take and so you should just wait for your turn for a non-timeboxed topic. In other words, the uncertainty of how long a discussion will need to take is very high once you get beyond an hour. But again, I don't mind much :)

template.md
bterlson: I don't really know what existing means so I was trying to add more rigor. Does existing simply mean it's been talked about in committee before? Had an actual agenda item before?

bterlson:
Added an agenda topic rules section with entries for timebox sizes and requirement for advancing proposals be added 7 days in advance. I suspect we will have to get actual consensus for this deadline before we can enforce it however :)

bterlson:
I've added a section for note-taking volunteers but in practice we don't use the same note takers for the entire time (and in general, the note taking process generally should be fixed!).

template.md
michaelficarra: I would say it has had an agenda item before. That means there was a discussion dedicated to that topic and we decided we'd like to continue the discussion in the future.

template.md
michaelficarra: I think those are both well reasoned.

michaelficarra:
I agree about fixing the note taking process. I would support a fund to get a dedicated note taker, if that's the route we want to take.

bterlson:
(Moving commit comment discussion from above to main thread)

Am fine changing "stage 0+ proposals" back to "existing proposals" where "existing" means "has been on the agenda before". Thinking about it more, though, I wonder if we can do better still by bucketing stage 1+ proposals looking to advance first, and reword "New proposals" as "Stage 0 proposals looking to advance". The stage 1+ bucket seems correct as any stage 1+ proposal has committee consensus on being something we should devote time to whereas a stage 0 proposal (whether it's been brought up before or not) doesn't have such consensus.

Another alternative is to simply have a "Proposals looking to advance" bucket and order that bucket by current stage (descending) followed by insertion order.

michaelficarra:
Both of those work for me.

littledan:
The agenda template looks great. Let's merge it and redo next meeting's agenda with it.

PR #195, opened 2016-06-21
Proposal to discuss nested `import` declarations

benjamn:
I'd like to present what I've learned from writing a new [transpiler](https://github.com/benjamn/reify) for ES2015 modules that does not restrict `import` declarations to the top level.

I think I can construct a convincing case for allowing nested `import` declarations, but I'm also sure I do not know all the relevant history that has gone into the current spec, which effectively restricts them to the top level.

Please have a look at the linked [document](https://github.com/benjamn/reify/blob/master/WHY_NEST_IMPORTS.md) before the July meeting (if this topic interests you), and let me know if it comes close to addressing your concerns. I'm happy to revise it or write a completely different sort of document, if folks prefer.

cc @dherman @caridy et al.


erights:
When I click on the "Merge pull request" button, it turns into "Confirm squash and merge". When I click on the down arrow on the left of this button is says

> Create a merge commit
> All commits from this branch will be added to the base branch via a merge commit.
> Not enabled for this repository
> ----------------------------
> Squash and merge

where the part above the separator is greyed out. Why?

benjamn:
@erights Same for me… seems like something @bterlson might know about?

michaelficarra:
It's fine. @bterlson doesn't like merge commits, so he has disabled creating merge commits from PRs in this and other TC39 repositories.

PR #198, opened 2016-07-07
Remove Unambiguous Module Grammar from July agenda.

jdalton:
Since [UnambiguousJavaScriptGrammar](https://github.com/bmeck/UnambiguousJavaScriptGrammar) was just accepted as [section 5.1](https://github.com/nodejs/node-eps/blob/master/002-es6-modules.md#51-determining-if-source-is-an-es-module) of the [Node ES6 Module Interoperability](https://github.com/nodejs/node-eps/blob/master/002-es6-modules.md) draft proposal I'm removing this item to give us time to get our feet wet, coordinate, and prepare a discussion at the Sept. TC39 meeting.


domenic:
So this will no longer be discussed at the TC39 meeting?

jdalton:
@domenic Correct. Not for the July meeting.

PR #201, opened 2016-07-26
Update 07.md

chicoxyzzy:


ljharb:
Thanks, but the current formatting is fine, and each proposal's entry should generally be edited by its champion.

mathiasbynens:
The way https://github.com/tc39/agendas/pull/201/files#diff-86b24a40af75c09871f3caa1cdd088acL56 gets rendered really makes it seem like it’s not supposed to be there though…

chicoxyzzy:
that's right

https://github.com/tc39/agendas/pull/201/files#diff-86b24a40af75c09871f3caa1cdd088acL61 is changed for consistency with other lines

PR #204, opened 2016-07-28
Slides for shared memory presentation

lars-t-hansen:


leobalter:
On my understanding, @allenwb suggested we should have the slides along with the meeting notes.

I don't have a preference I would die for, but I believe we should have a final and consistent decision on this. One single place is batter than many.

rwaldron:
Sorry, this format is not acceptable. Slides that aren't hosted on some web platform need to be provided as PDF _only_.

ljharb:
What is ODP?

leobalter:
Open Document Presentation, used w/ OpenOffice

rwaldron:
For context, slides are entered into the official record with Ecma

lars-t-hansen:
I've added a PDF, I'll submit a PR to fix the link.

rwaldron:
Thank you

PR #208, opened 2016-07-30
Add RegExp Unicode property escapes proposal to September agenda

mathiasbynens:
Champions: @littledan @bterlson


ljharb:
LGTM, but I'll let the champions merge it.

PR #215, opened 2016-09-05
Set, Map, WeakSet, and WeakMap: `of` and `from` methods

leobalter:


ljharb:
Added to the proposals list in https://github.com/tc39/proposals/pull/21

PR #225, opened 2017-01-18
Use the Covenant CoC only

leobalter:
Ref c7796e0d43d3e8f6afd5cd01b570b40267190219 (see the comments)

My point is to defer to other CoCs only if this one is rejected.

I'm also time-boxing this so if we can't reach consensus before 30 minutes I will consider other forms.


leobalter:
cc @littledan

littledan:
SGTM

PR #239, opened 2017-05-24
add slides

hzoo:
Should I link to slides directly or the repo?

> https://github.com/hzoo/role-of-babel-in-js
> http://henryzoo.com/role-of-babel-in-js/

@littledan


kentcdodds:
Both would be good. But actually probably the slides are better.

PR #247, opened 2017-07-19
Added proposals for discussion

rbuckton:


domenic:
This is not appropriate:

> Proposals looking to advance must be added to the agenda along with necessary review materials 7 days prior to the first day of the meeting.

rbuckton:
My apologies. I've been trying to discuss with @bterlson for the past few days but he's been busy. I may not be able to attend the next few meetings. As this meeting is being held in Redmond, it may be my only opportunity to speak on these for some time.

domenic:
That's OK; sorry for being a bit terse. I'm a bit frustrated since we just had our internal meeting to discuss the agenda, on the assumption that people would be following the rules.

I think the best course of action is to remove the item seeking stage advancement, as that goes against the process. New proposals not seeking to advance are fine to discuss.

bterlson:
This proposal is unchanged from the previous meeting where it was on the agenda. It should have been carried forward from the previous agenda as we did not get enough time to talk about it then. I think this is fine.

rbuckton:
I was looking to discuss advancing Cancellation as we were cut short on time in the last meeting. It's primarily aligned with "9. Timeboxed overflow from previous meeting".

domenic:
I still think this is not appropriate, but if it is unchanged from the previous meeting, then it is a moot point, as the objection to advancement in its form-as-of-then will stand. I don't think it's a great use of committee time to reiterate that, but with an appropriate timebox, maybe it'll be OK.

ljharb:
Sounds like this is fine to merge? I'll do so shortly if there's no objections.

bterlson:
I don't think someone who wants to litigate my editorial choices during plenary gets to decide what is and is not a good use of our time :-P

I also note that the notes do not contain any objections, just areas of further investigation and questions. If you intend to block this proposal you should be on record doing so (and with good reasons for posterity)!

I think this is good to merge, though would be ok moving it to overflow from previous if people object to its current placement.

rbuckton:
The main issue was that there was not time to fully understand the objection to advancement. I'd like to continue the discussion where we left off, even if we timebox it to 30 minutes (which is the remainder of the time from last meeting).

domenic:
Sorry, I guess the note-takers were incomplete. We definitely intend to block this proposal in its current form from advancement, and gave concrete reasons in plenary; I am sorry that was not recorded.

I object to the stage-advancement item being added this late, but I'm not sure whether the agenda pull requests is the right forum to resolve that, and am willing to add a timebox to get us past that. I still would prefer to discuss the new-proposal items in preference to one that hasn't changed, however.

leobalter:
I'd be fine to add the items to the agenda until we formalize the process to freeze it in time.

The objection to advancement in its form might stand and it should be respected, but being fair with every delegate, this is not the only time we have something being added late.

I'm +1 to merge this for now, as I'll raise the formality issue in general during the meeting next week.

rbuckton:
Should I move Cancellation up to the overflow section?

ljharb:
@rbuckton yes, let's move it to overflow since that's what it is, and then i'll merge this - we can address further issues during the meeting.

leobalter:
> Should I move Cancellation up to the overflow section?

I believe it's reasonable and fair.

rbuckton:
I'm uncertain as to whether this should go under 9.iv. as it *was* a 60 minute item, or 9.ii. as we only spent 30 minutes on it in May.

ljharb:
@rbuckton the new timebox is up to you; you can pick whatever time makes sense.

rwaldron:
> Sorry, I guess the notes taken were incomplete.

@Domenic this is why you're given a full week of time to review the notes and make important additions or changes. In the future, I encourage you to take advantage of that period.

littledan:
This isn't about the notes. I think it'd be nice for delegates who represent larger organizations to have time to prepare based on the agenda. Someone could say at the previous meeting, "I'll present on this next time", and then not end up presenting it. It's also useful to have linked documents available in advance. For example, Googlers may want to talk with other people doing web platform work on cancellation who don't tend to come to TC39.

The takeaway here can be, when in doubt, if you're considering something, put it on the agenda at the risk of taking it off later, rather than delay in putting it on. I've done that with two topics where I'm not completely sure whether I'll present them or delete them later (pipeline and literals).

Maybe we could do a better job communicating these goals before adopting a punitive stance of not discussing the things added at the last minute, though. For now, most of these new items are already at the bottom of the agenda.

PR #256, opened 2017-09-19
Add "referential destructuring" proposal to 30min-boxed items.

benjamn:
This is a new (Stage 0) proposal of a feature that should improve the usability of dynamic `import()`, and whose adoption would persuade me to abandon my old [nested `import` declarations](https://github.com/tc39/ecma262/pull/646) proposal.

Please see the linked [proposal text](https://github.com/meteor/proposal-referential-destructuring/blob/master/README.md) for more information. I believe it meets the requirements for Stage 1.

### DIFF
--- a/2017/09.md
+++ b/2017/09.md
@@ -102,6 +102,7 @@ Allen's paper on standards committee participation for new attendees: http://wir
         1. [Throw Expressions](https://github.com/rbuckton/proposal-throw-expressions) for Stage 2 (Ron Buckton) ([slides](https://rbuckton.github.io/proposal-throw-expressions/ThrowExpressions-tc39.pptx), [spec](https://rbuckton.github.io/proposal-throw-expressions/))
         1. [`do`-expressions](https://github.com/tc39/proposal-do-expressions) update (Dave Herman)
         1. [Ergonomic Cancellation Through Async Calls](https://docs.google.com/presentation/d/1Lg_chkGiugx69yXKwr5Vrg4Xmxt7Tui0kmUFFZx1KNs/edit?usp=sharing) (Yehuda Katz, David Herman)
+        1. [Referential destructuring](https://github.com/meteor/proposal-referential-destructuring) for Stage 1 (Ben Newman)
     1. 45 Minute Items
         1. [String.prototype.matchAll](https://github.com/tc39/proposal-string-matchall/pull/17) for Stage 2 (Jordan Harband)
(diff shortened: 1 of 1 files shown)


benjamn:
Sorry, I thought submission by today would be sufficient! Happy to defer to other agenda items that were added before this.

domenic:
I just want to say thanks to @ljharb for helping enforce the norm we're attempting here. It makes it much easier for those of us trying to prepare for the meeting.

littledan:
Agree that deadlines are good, but I also thought today was the deadline. If it should be the Monday before, maybe we can be more unambiguous about this next time. How about, for next meeting, we write the deadline date on the agenda itself?

ljharb:
I'm fine either way; either this PR is after the deadline, or it's the final PR before the deadline - we just need consensus on which. Regardless, it's on the bottom of the priority list for presenting; the only real impact of "which" is whether it's eligible for stage advancement in September or not. I don't have a strong opinion either way, but given the ambiguity, I wanted to call it out.

Moving forward, we should decide precisely when the cutoff is; I'd suggest 10AM (meeting start time), 7 days prior to the first day of the meeting, based on the timezone where the meeting occurs, to avoid future ambiguity.

benjamn:
I agree that 10am Tuesday a week before the meeting is a defensible hard deadline, if stated explicitly in the agenda.

Realistically, I doubt we will have time for me to present this proposal, given how packed the agenda already is, and that’s ok. Mostly hoping to put the idea on folks’ radars, as it might have serious flaws or weird interactions with other destructuring proposals, such as @rbuckton’s.

leobalter:
We should definitely have a better process to freeze the agenda in a specific time before the meeting. Although, considering the nature of other proposals and meetings I'm willing to allow a separate queue for new items that can still be evaluated __if we find available time__ in the meetings.

The consequences might be naturally someone saying they didn't have enough time to review it or just update the committee with useful information. I'd support both.

At the same time, it would be great if more people could be reviewing the agenda items in time for the meetings. I frequently see some any questions there could be answered by a simple proposal review. A precious time use could be improved by more async efforts from delegates.

---

I think it's still worth to register items added after the freezing time. These shouldn't not go as priorities for the next agenda, but still helps listing what should be considered. Maybe I'm overthinking this and we should just add these to the next agenda already.

@benjamn do you want to update this PR and move the item to the next meeting?

benjamn:
@leobalter As a New Yorker and a small-company delegate, I'm not sure I'll be able to make the next meeting in SF (definitely going to Boston, though), so I'd like to keep the item on this agenda, even if there isn't enough time for me to present. At least people who might be interested in the proposal will see it, and I can discuss it with them personally in Boston.

ljharb:
Sounds good; I'll merge this now, and I'll update the agenda to note the items landed at the last minute.

PR #257, opened 2017-09-20
Add ⌛️ to denote late items

ljharb:


leobalter:
I just saw @ljharb also requested a review from @littledan and @bterlson.

I believe it's fine for now and we can fix or improve this in follow up PRs.

syg:
Wait, asyncWait was added 7 days before (though, granted, exactly 7). What makes it late?

benjamn:
Neither of these proposals are "late" according to any stated criteria (as of Tuesday), but I've decided to refrain from arguing this point unless advancement seems like a viable possibility. That said, you did insert your proposal above other proposals in the list (without a PR)—any particular reason for that? Do we have any process for deciding the order of the items?

syg:
I had thought all the Need Consensus PR were grouped. They tend to get their own time period and are prioritized, usually.

-- 
       shu

ljharb:
@syg I believe yours was a few hours past the Boston-timezone start time of Tuesday's meeting.

Needs Consensus PRs are all indeed grouped.

syg:
I see. “Days” to be interpreted as 24 hours prior to start time in host time zone was not at all clear to me, nor do I think is a widely shared understanding. I am happy to be at the bottom of the queue. I would like to voice my opinion that this is kind of  ridiculous. I’m sure everyone would appreciate a clearly spelled out cutoff time going forward.

-- 
       shu

littledan:
I'm all for marking late items, but I'm not sure if these items deserve to be marked late. For now, I'd interpret 7 days in a slightly more loose way, to account for the ambiguity the first time. I remember saying a week; I don't remember stating a particular time of day. As long as something was put on the agenda by the Tuesday before the meeting started, I think that's enough given what was discussed before. For next time, we can be more explicit about the exact time, as proposed in another thread.

I put needs-consensus PRs on the agenda because in the past we ended up skipping them. I wanted to timebox them, and in the past, when I tried to put timeboxed items on the agenda under other items, people objected, so I put them under the main timeboxed groups.

ljharb:
See #259 for the thrilling consensus-driven conclusion

PR #259, opened 2017-09-20
Clarify hourglass rules

### LINKED ISSUE #258: Proposal: freeze the agenda in the second Friday before meetings
This a process proposal that shouldn't be effective for the meeting of 2017-09.

Freezing the agenda in the second-to-last Friday before the meetings might be helpful. Things to consider in this case:

- Friday could be considered within each one's timezone, regardless of location.
- This still provides others a full week, Monday to Friday, to review items before the meeting, regardless of timezone/location and the time required for traveling.
- We can allow some flexibility for items added a bit late. A Friday night commit in the Pacific time wouldn't affect the available week days to review the work.
- It gives us some extra few days compared to the current policy.

If we are positive with this change, we should have this explicitly set as a policy for the next agendas.

ljharb:
Per discussion on https://github.com/tc39/agendas/commit/950e2410cf3d60eebc367f986b4b9b70dd669f7b, and in the 2017.09.28 TC39 meeting.

Closes #258.

### DIFF
--- a/2017/11.md
+++ b/2017/11.md
@@ -34,8 +34,12 @@ TBD Hotels nearby (optional)
 
 ## Agenda Topic Rules
 
-1. Proposals looking to advance must be added to the agenda along with necessary review materials 7 days prior to the first day of the meeting.
-   1. Note: any items added after 7 days prior to the start time of the meeting must be denoted with an ️:hourglass: symbol, which indicates that it is not eligible for stage advancement, and that it must wait until all other items have been addressed.
+Deadline for advancement eligibility: **November 18th, 2017, 18:00 UTC**
+
+1. Proposals not looking to advance may be added at any time; if after the deadline, please always use a pull request so that members are notified of changes.
+1. Proposals looking to advance to stage 0 or 1 must be added (and noted as such) prior to the deadline.
+1. Proposals looking to advance to stages 2, 3, or 4 must be added (and noted as such) *along with the necessary materials* prior to the deadline.
+    1. Necessary materials may include slides, a link to the proposal repo, a link to spec text, etc.
 1. Timeboxed topics may be 15, 30, 45, or 60 minutes in length.
 
 ## Agenda items
--- a/2017/11.md
+++ b/2017/11.md
@@ -34,13 +34,17 @@ TBD Hotels nearby (optional)
 
 ## Agenda Topic Rules
 
-Deadline for advancement eligibility: **November 18th, 2017, 18:00 UTC**
+Deadline for advancement eligibility: [**November 18th, 2017, 18:00 UTC**](https://www.timeanddate.com/countdown/generic?p0=1440&iso=20171118T18&msg=TC39%20Submission%20deadline)
+<sub>Note: this time is selected to be precisely 10 days prior to the start of the meeting</sub>
 
 1. Proposals not looking to advance may be added at any time; if after the deadline, please always use a pull request so that members are notified of changes.
 1. Proposals looking to advance to stage 0 or 1 must be added (and noted as such) prior to the deadline.
+    1. Such proposals *should* include supporting materials when possible.
 1. Proposals looking to advance to stages 2, 3, or 4 must be added (and noted as such) *along with the necessary materials* prior to the deadline.
-    1. Necessary materials may include slides, a link to the proposal repo, a link to spec text, etc.
+    1. Such proposals *must* link to supporting materials prior to the deadline, or they will not be eligible for advancement. If these materials change substantially after the deadline, the proposal may be disqualified from advancement, based on the committee’s judgment.
 1. Timeboxed topics may be 15, 30, 45, or 60 minutes in length.
+
+Supporting materials includes slides, a link to the proposal repository, a link to spec text, etc.; essentially, anything you are planning to present to the committee, or that would be useful for delegates to review.
 
 ## Agenda items
 
--- a/2017/11.md
+++ b/2017/11.md
@@ -35,7 +35,7 @@ TBD Hotels nearby (optional)
 ## Agenda Topic Rules
 
 Deadline for advancement eligibility: [**November 18th, 2017, 18:00 UTC**](https://www.timeanddate.com/countdown/generic?p0=1440&iso=20171118T18&msg=TC39%20Submission%20deadline)
-<sub>Note: this time is selected to be precisely 10 days prior to the start of the meeting</sub>
+  - <sub>Note: this time is selected to be precisely 10 days prior to the start of the meeting</sub>
 
 1. Proposals not looking to advance may be added at any time; if after the deadline, please always use a pull request so that members are notified of changes.
 1. Proposals looking to advance to stage 0 or 1 must be added (and noted as such) prior to the deadline.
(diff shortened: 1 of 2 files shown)


template.md
```diff
@@ -34,8 +34,8 @@ TBD Hotels nearby (optional)
 
 ## Agenda Topic Rules
 
-1. Proposals looking to advance must be added to the agenda along with necessary review materials 7 days prior to the first day of the meeting.
-   1. Note: any items added after 7 days prior to the start time of the meeting must be denoted with a ⌛️ symbol, which indicates that it is not eligible for stage advancement, and that it must wait until all other items have been addressed.
+1. Proposals looking to advance must be added to the agenda (meaning, a PR opened and/or merged) along with necessary review materials 7 days (168 hours, precisely) prior to the first day of the meeting.
```
leobalter: _I can imagine we should use moment.js at some point to count how many hours left._

`s/'a PR opened and/or merged'/'a PR opened and/or a commit merged'`

this case covers direct commits even if they lack our full appreciation.

ljharb: Good point, will update

littledan:
Maybe we should hold off on this in favor of @leobalter 's proposal.

ljharb:
This PR has now been updated with the rules we agreed to today.

2017/11.md
```diff
@@ -34,8 +34,12 @@ TBD Hotels nearby (optional)
 
 ## Agenda Topic Rules
 
-1. Proposals looking to advance must be added to the agenda along with necessary review materials 7 days prior to the first day of the meeting.
-   1. Note: any items added after 7 days prior to the start time of the meeting must be denoted with an ️:hourglass: symbol, which indicates that it is not eligible for stage advancement, and that it must wait until all other items have been addressed.
+Deadline for advancement eligibility: **November 18th, 2017, 18:00 UTC**
```
domenic: Link to https://www.timeanddate.com/worldclock/fixedtime.html?msg=TC39+Submission+deadline&iso=20171118T18&p1=1440 or https://www.timeanddate.com/countdown/generic?p0=1440&iso=20171118T18&msg=TC39%20Submission%20deadline ?

2017/11.md
```diff
@@ -34,8 +34,12 @@ TBD Hotels nearby (optional)
...
+
+1. Proposals not looking to advance may be added at any time; if after the deadline, please always use a pull request so that members are notified of changes.
+1. Proposals looking to advance to stage 0 or 1 must be added (and noted as such) prior to the deadline.
```
domenic: "Such proposals _should_ include supporting materials when possible"

2017/11.md
```diff
@@ -34,8 +34,12 @@ TBD Hotels nearby (optional)
...
+1. Proposals looking to advance to stages 2, 3, or 4 must be added (and noted as such) *along with the necessary materials* prior to the deadline.
```
domenic: "Such proposals _must_ link to supporting materials prior to the deadline, or they will not be eligible for advancement. If these materials change substantially after the deadline, the proposal _may_ be disqualified from advancement, based on the committee's judgment."

2017/11.md
```diff
@@ -34,8 +34,12 @@ TBD Hotels nearby (optional)
...
+    1. Necessary materials may include slides, a link to the proposal repo, a link to spec text, etc.
```
domenic: Move this after the list of rules and change it to something like "Supporting materials includes slides, a link to the proposal repository, a link to spec text, etc.; essentially, anything you are planning to present to the committee, or that would be useful for delegates to review."

2017/11.md
ljharb: Nice idea, I like the countdown

ljharb:
@littledan @leobalter @bterlson i'd prefer to have stamps from at least all three of you before merging this :-) the more reviews on it, the better!

leobalter:
LGTM, but might be good to specify when the deadline should be set, the current text seems not clear. Correct me if the information is there and I missed it.

2017/11.md
```diff
@@ -34,10 +34,17 @@ TBD Hotels nearby (optional)
 
 ## Agenda Topic Rules
 
-1. Proposals looking to advance must be added to the agenda along with necessary review materials 7 days prior to the first day of the meeting.
-   1. Note: any items added after 7 days prior to the start time of the meeting must be denoted with an ️:hourglass: symbol, which indicates that it is not eligible for stage advancement, and that it must wait until all other items have been addressed.
+Deadline for advancement eligibility: [**November 18th, 2017, 18:00 UTC**](https://www.timeanddate.com/countdown/generic?p0=1440&iso=20171118T18&msg=TC39%20Submission%20deadline)
```
ljharb: @leobalter this line specifies the precise time for the deadline; if you mean specifying the rubric (10 days prior) that's a great idea, and I can add that in :-)

leobalter:
Yes, as rule for the template. Thanks

ljharb:
@leobalter updated!

2017/11.md
```diff
@@ -34,10 +34,18 @@ TBD Hotels nearby (optional)
...
+  - <sub>Note: this time is selected to be precisely 10 days prior to the start of the meeting</sub>
```
leobalter: The main goal is to have it defined with a strict precision. I don't want to block this anymore but I will write a follow up for mention it should also be a UTC -12 definition.

leobalter:
> leobalter approved these changes on behalf of tc39/delegates just now

This github feature is confusing.

ljharb:
@leobalter i was still hoping to have rex review this first; in general, it's common etiquette when the PR author has commit access that they be the one to do the merge.

2017/11.md
ljharb: My intention was for it to remain blocked until it got as close to 100% of delegate approvals as possible.

leobalter:
We share different common etiquettes, where I usually don't merge my own submitted PRs.

It might have seem that I jumped the trigger but I believe the current text is reasonable, also working on a small follow up (see #264), helps the process with smaller but specific topics to address.

RexJaeschke:
I am OK with the direction you all are heading.

2017/11.md
littledan: More time for reviews is always helpful, and later means easier to contribute. This is a reasonable arbitrary point in the middle. Not sure why there's more need for rationale text here than elsewhere in the agenda.

littledan:
Sorry for not reviewing earlier. A couple suggestions:
- The rules have pretty imposing wording. I'd prefer something more friendly and inviting. We aren't trying to keep anybody out, just to share their proposal ahead of time to maximize the feedback they can get.
- Now that the rules are so long, would it make sense to break then out into a separate page?

ljharb:
Sure, do you have any specific suggestions for making it more friendly?

I don't think they're that long; just a couple lines - i feel that having them inline makes them more likely to be read, and harder to be ignored.

littledan:
For friendly wording, things could be phrased in terms of things like "It's helpful to the committee members if you do this..." rather than necessarily *should* in italics. For the "must" items, the wording could be more like, "unfortunately, we won't be able to review things that are submitted this way."

For slides as a supporting document, I don't think that many committee members currently prepare slides ahead of time. Although I sometimes write slides further in advance, when I write slides, I write them to be accompanied by talking, not to be a standalone document that actually describes the proposal. I think committee members should be encouraged to put the real content in explainer documents or spec text, with the slides as a small accompanying thing that doesn't necessarily need to be made available ahead of time.

ljharb:
Both seem reasonable; I'll write up a PR for those changes soon.

domenic:
I find "should" and "must" useful because I know what they mean via https://tools.ietf.org/html/rfc2119 , FWIW.

ljharb:
The spec definitions of should/must are what I inferred from it as well; I don't feel strongly either way. We can hash it out on the PR :-)

PR #260, opened 2017-09-25
Remove second listing of iterator change PR

shvaikalesh:


littledan:
I think we should leave in the first of the two lines you deleted, in case Brian wants to present any additional PRs (as he usually does during that item).

2017/09.md
```diff
@@ -124,7 +124,6 @@ Allen's paper on standards committee participation for new attendees: http://wir
     1. [Pattern Matching Updates](https://github.com/tc39/proposal-pattern-matching) (Brian Terlson)
 1. Non-timeboxed agenda items
     1. Web compatibility issues / Needs-consensus PRs
-      1. https://github.com/tc39/ecma262/pull/988
```
ljharb: The "Needs Consensus" section is the proper place for PRs; if they exist elsewhere, either the duplicates should remain, or they should be removed - but it definitely needs to be here.

littledan: I don't understand the purpose of duplicating them. With or without this patch, we don't have an accurate duplicated section. How about we leave this for any additional PRs which aren't discussed earlier, since the partial duplication was confusing (at least to me)?

littledan:
Merging based on approval from Elkhart above.

PR #262, opened 2017-09-27
Apply the Herman Normalization algorithm

chancancode:
cc @dherman


dherman:
I believe the technical definition of dherman-normal form involves my social security number, which I can't disclose for security reasons.

/cc @equifax for review

leobalter:
I wonder if the former version is covered by Annex B.

PR #264, opened 2017-10-03
Fix UTC-12 time for deadline

leobalter:


2017/11.md
```diff
@@ -34,8 +34,8 @@ TBD Hotels nearby (optional)
 
 ## Agenda Topic Rules
 
-Deadline for advancement eligibility: [**November 18th, 2017, 18:00 UTC**](https://www.timeanddate.com/countdown/generic?p0=1440&iso=20171118T18&msg=TC39%20Submission%20deadline)
-  - <sub>Note: this time is selected to be precisely 10 days prior to the start of the meeting</sub>
+Deadline for advancement eligibility: [**November 17th, 2017, 23:59:59 UTC-12**](https://www.timeanddate.com/countdown/generic?iso=20171117T235959&p0=3926&msg=TC39+Submission+deadline&ud=1&font=cursive)
```
ljharb: I explicitly chose 10 AM PST as the deadline, so it would match the start time of the meeting; this also corresponds more naturally to a comfortable time in the timezone of the meeting.

The consensus was to pick a specific time; nobody cared too much what the time was.

leobalter: I picked the suggestion from @domenic in https://github.com/tc39/agendas/issues/258, which is the same I proposed during the meeting.

The problem with the deadline being flexible is the main point to solve the pain point here. 

Also part of the consensus is:

> - Specify an exact time for each meeting

leobalter: > Also part of the consensus is ...

@ljharb you mentioned that, sorry for just repeating.

ljharb: Is there a reason that the extra 18 hours makes a difference? I don't think "midnight UTC" is something that's any more intuitive than anything else, and whatever time we pick, the countdown link makes it unambiguous, and easily convertible into any timezone.

leobalter: the flexibility between each meeting causes confusion, also the UTC -12 guarantees everyone will have the whole Friday as a deadline, regardless of their location.

ljharb: This time is Friday at 11:59PM UTC, which is 5PM PST - in other words, it ensures people have the whole Saturday, but only allows a partial Friday. If you want to ensure the whole Friday, we'd have to do UTC midnight *eleven* days before the meeting.

maggiepint: I think this is the formula you're describing, where the date there can be replaced with the meeting date:
```js
moment.tz('2017-11-28','Etc/GMT+12').day(-2).subtract(7,'days').endOf('day').utc().format()
"2017-11-18T11:59:59Z"
```

This is basically, take the meeting date in UTC-12, set it to the last Friday, take off 7 days before that, go to the end of that local day, convert back to UTC, and format out date.

littledan:
Given that this deadline change didn't land, and the current deadline has an unambiguous UTC time listed on it (regardless of whether it was chosen in a good way), let's use the current thing in master. Having this PR open creates some ambiguity. We can discuss a change in policy for the next meeting later, in a PR against its agenda.

PR #266, opened 2017-10-11
Update with Rex’s suggested changes

ljharb:
The only item that seems like it maybe could have some discussion is moving the "needs consensus" PRs up to the top, and adding the timebox sections to them.

I'll plan to merge this in a day or so if there's no objections.


littledan:
LGTM

PR #268, opened 2017-10-18
Array Security Presentation

natashenka:
Signing up to give a quick presentation on vulnerabilities in Array[@@species] and Array index accessors
@ajklein


leobalter:
I wonder if this item will demand more than 15 minutes for discussion. I'm interested, and I believe other people will be as well. Maybe 30 minutes should be better for this case?

michaelficarra:
@natashenka If you could link to any materials we could use to prepare, that would also help with timing. If you're concerned about publicising details, you can pass the materials along to someone who will safely distribute them to delegates via our Reflector.

natashenka:
You can see my slides here (if you can't access them, request permission and I'll respond.) : https://docs.google.com/a/google.com/presentation/d/1t8SX0MP4-PswZnKPBMm7yaCfBcmWR_QfIc4LWh9Fydc/edit?usp=sharing . Nothing confidential, but not quite ready for prime time. Thinking about it a bit more, I think 30 minutes might be better including the discussion part. Let me know what you think after reading them.

natashenka:
So, after seeing the materials, do you think 15 minutes is appropriate? Or should I move it to the 30 minute section?

ljharb:
It’s better to ask for 30 and need 15 than the reverse :-) probably best to move it just in case.

ajklein:
30 minutes sounds more appropriate to me, too.

PR #273, opened 2017-11-10
Add item to agenda regarding interpreter directives

bmeck:


michaelficarra:
@bmeck since I don't think mentions in gists trigger notifications, linking you my comment so you can see it before the meeting: https://gist.github.com/bmeck/59cf8c16959eccffd8b7e9828826a842#gistcomment-2254745

bmeck:
@michaelficarra moved to HashBang since throwing back to thoughts of https://github.com/rwaldron/tc39-notes/blob/master/es7/2016-09/sept-29.md#for-decorator + `She-` prefix isn't what I think of when I see `#`

PR #276, opened 2017-11-17
Add string literal distinction to the agenda.

mikewest:
@mathiasbynens, @ajklein: WDYT? I'm not really sure this deserves 60 minutes of folks' time, and I'd be fine with timeboxing it to something smaller. Y'all know the group (much) better than I do, so I'd appreciate your advice. :)


littledan:
Looks good to me! I think this topic is important enough to spend a long time discussing, but I'm a little worried that we could get off-track with people objecting at length to details of something you're not even proposing. It could help things go smoothly if you do one of two things:
- Come up with some strawman details that fit well with the rest of the language that you'd be confident defending. I outlined one possibility at https://github.com/mikewest/tc39-proposal-literals/issues/2
- Scope the discussion at the beginning to understanding the context and motivation, being very explicit that you're not proposing that these specific details get into the language. Further development of details might happen at the meeting, or might be more productive either offline on GitHub or in a follow-up meeting. For this, a shorter timebox might be helpful, or might not be if there's enough to present at a high level for 60 minutes.

mikewest:
Thanks for the feedback, Daniel!

I appreciate your points about narrowing the focus to the context and motivation. I honestly don't have enough context with the innards of the language to make defensible proposals. Instead, I have use cases, and vague sketches of what I'd like to see from the platform side, and I'd love to get the group's help on the details of what the language might be willing and able to provide in order to meet those.

How would you recommend restructuring the proposal to make that clear?

littledan:
That sounds like a good phrasing. If you make that clear as you're presenting it, it will be helpful.

PR #279, opened 2017-11-18
object.freeze+object.seal syntax to 30 minute items

keithamus:


keithamus:
I see theres some contention of the cutoff time in #264 but given the deadline in `master` is still 5 and a half hours away I feel justified in raising this at this time.

Issue #271, opened 2017-11-07
I have too many agenda items


littledan:
I don't want to crowd out other people from talking, or take over topics that other want to champion. However, I want to help some important topics move forward. For the balance, maybe a couple things could be adjusted in the agenda, but I am wondering about your thoughts:
- I could deprioritize my topics in the agenda (within the category) to give everyone else a chance to speak first
- I can mention that I'd be happy to have others presenting these topics if they want to (but I am not sure when in the meeting that would be appropriate).

Any other ideas? We have a general issue whhere there are a lot of smart people in the room who don't seem to be empowered to champion proposals.

snuggs:
@littledan I run NYC.rb and is a meetup group of about 5,000 members. Whenever we are obtaining consensus during organizer meetings we often limit questions to one per person (amongst other conventions). This usually cuts down at least 40-50% of deliberation time.

Also have read some interesting protocols in [Robert's Rules of Order](https://en.wikipedia.org/wiki/Robert%27s_Rules_of_Order)

TL;DR; _Even if an organization has adopted Robert's Rules of Order, it can still adopt its own rules which supersede any rules in this book._

YMMV of course and take it as a grain of salt. At minimum guaranteed to help break through a cognitive stalemate as an organizer.

jridgewell:
@snuggs: Not quite a perfect fit. @littledan _presents_ several (too many apparently) proposals during the meetings, he doesn't hog up discussion time after the presentation.

bakkot:
@littledan I don't personally have any problem with you presenting on a bunch of different topics. It's better than them being neglected!

That said, I'd be happy to present the subclassing-private-static item you have on the agenda for this meeting, especially if you don't mind me just using the slides you've already made.

snuggs:
@jridgewell Ah i see. Just sounds like a delegation situation then. In that case I'd say if I can't get at least one person that can cheerlead the proposal I probably am going to have a difficult time gaining interest in the first place. Is there anyone that can champion some?

snuggs:
Oh right! What @bakkot said... Github doesn't seem to update comments as frequently as you'd think. #HardRefreshLife :-/

littledan:
@bakkot Sounds good. I gave you edit access to the slides and wrote your name next to the agenda item. Thanks for your help!

littledan:
@snuggs Interesting idea!

keithamus:
Can this be closed now @littledan has moved their topics to the end?

[...]

PR #280, opened 2017-11-18
Move my topics to the end of their sections

littledan:
I'm presenting many topics. To encourage others to present and not hog all the time, this puts the timeboxed items where I was going to present at the bottom of their sections. Does such a change seem OK?

Addresses #271

### DIFF
--- a/2017/11.md
+++ b/2017/11.md
@@ -86,34 +86,34 @@ Supporting materials includes slides, a link to the proposal repository, a link 
         1. [Numeric separators](https://github.com/tc39/proposal-numeric-separator) for Stage 3 (Sam Goto, Rick Waldron)
         1. [RegExp named captures](https://github.com/tc39/proposal-regexp-named-groups) for Stage 4 (Mathias Bynens)
         1. [RegExp lookbehind assertions](https://github.com/tc39/proposal-regexp-lookbehind) for Stage 4 (Mathias Bynens)
-        1. [BigInt](https://github.com/tc39/proposal-bigint) status update (Daniel Ehrenberg) ([slides](https://docs.google.com/presentation/d/1u2xXRokUBPMjBsTL_ZCDghA16cH5FI0m1kLOczm1oMw/edit#slide=id.p))
-        1. [Intl.ListFormat](https://github.com/tc39-transfer/proposal-intl-list-format) find Stage 3 reviewers (Daniel Ehrenberg)
         1. InterpreterDirective (Bradley Farias) ([gist](https://gist.github.com/bmeck/59cf8c16959eccffd8b7e9828826a842))
         1. [`String.prototype.codePoints`](https://github.com/RReverser/string-prototype-codepoints) for Stage 1 (Ingvar Stepanyan, Mathias Bynens)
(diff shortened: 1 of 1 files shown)


2017/11.md
```diff
@@ -86,34 +86,34 @@ Supporting materials includes slides, a link to the proposal repository, a link
...
-        1. [Allow await in non-arrow async function parameter defaults](https://github.com/tc39/ecma262/issues/917) (Daniel Ehrenberg)
         1. Revisiting parameter initializers and sloppy eval (Adam Klein) ([gist](https://gist.github.com/ajklein/b947351835cc77ad0040db9a55813f51))
         1. Discuss module order instantiation/evaluation guarantees (Bradley Farias) ([slides](https://docs.google.com/presentation/d/1RXvvScD8ce2FyLY2aYhbas83WCiBqzIOqdMt4OpkCJM/view))
         1. Repair loss of Proxy transparency, for Stage 1 (Mark S. Miller, Caridy Patino, Keith Miller, Tom Van Cutsem) ([issue thread](https://github.com/tvcutsem/es-lab/issues/21))
         1. Array`[@@Species]`, Array Index Accessors and Security (Natalie Silvanovich)
+        1. [Intl.Locale](https://github.com/tc39/proposal-intl-locale) for Stage 2 (Daniel Ehrenberg)
+        1. Clarify/redefine Stage 4 requirements (Daniel Ehrenberg) ([PR](https://github.com/tc39/process-document/pull/15))
```
mathiasbynens: I think this topic should be discussed first, as other agenda items depend on its resolution.

littledan: Would first among 30-minute items be enough, or does it need to be even further forward?

mathiasbynens: Even further IMHO. It should be one of the first items of the entire meeting.

littledan: @ljharb How would you feel about that?

ljharb: Seems appropriate to me; having it jump the queue seems like the sort of thing @RexJaeschke should/would decide.

kmiller68: The stage 4 requirements are of particular interest to me. Would it be possible to move the discussion to after noon on the first day? Due to traveling constraints, I will, unfortunately, be late the first day.

littledan: @ljharb Currently, the chair group works as a team; it's not one of us making all the decisions. Initially, Rex was proposing that the three of us meet ahead of time to decide on agenda priorities and ordering. However, feedback from some on the committee (including @ljharb) was that we should stick with the rules-based approach. So, as such, we aren't organizing those pre-meetings.

Given that, I think we should discuss here on GitHub whether it's OK to do such a reordering. If you need chair group signoff, I can provide that; I wouldn't expect others to be opposed to this either.

ljharb: Good point :-)

I think it's totally OK; in past meetings we frequently reorder things as needed for scheduling constraints, or because a topic depends on the result of another topic.

This one affects anything in stage 3 or 4, so it both makes sense to prioritize it, and also I can't imagine anyone would object to discussing it first.

PR #286, opened 2017-12-15
More info about the January meeting

samuelgoto:


samuelgoto:
Staged here:

https://github.com/samuelgoto/agendas/blob/master/2018/01.md

PR #288, opened 2018-01-09
missing overflow for standard library

bmeck:


littledan:
Did you consider adding a timebox to this topic, and then it could be discussed earlier in the meeting?

bmeck:
@littledan as an overflow item that was not originally timeboxed and my not being there in person, I did not.

PR #296, opened 2018-01-12
Add 30 minute time slot to discuss Top Level Await

MylesBorins:
I'd like to have a 30 minute time box discussion about bringing `Top Level Await` to stage 0.

I've had a combination of vacation + illness get in the way of me getting any presentation material together yet, but am confident I can have something together in time for the meeting. Please let me know if that is acceptable. If not we can push until March


littledan:
If you're not going for stage advancement beyond Stage 1, the rules at the top of the agenda say it's acceptable to publish presentation materials later.

PR #298, opened 2018-01-13
Add agenda list to the front page of the readme

ljharb:
This commit both adds a list of all agendas to the front page of the readme, and also adds `./scripts/generate.js` for future updates.

I intentionally did not put `package.json` in the root of the repo (along with `.npmrc`) to try to keep the web view of the root clean; by doing so, it would enable something simple and discoverable like `npm run build`. However, this script needs to be run basically once per new agenda, so, 6 times per year, and doesn't need to be that discoverable - so I'm not sure if it matters.

(This can be previewed [here](https://github.com/ljharb/agendas/tree/toc))

Is this a good idea? What do people think?


scripts/package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "private": true,
+  "devDependencies": {
+    "dedent": "^0.7.0",
+    "glob": "^7.1.2",
+    "markdown-magic": "^0.1.20"
+  }
```
mathiasbynens: Let’s add a `scripts` section with a `build` command so people can run `npm run build` instead of `node scripts/generate.js`.

scripts/generate.js
```diff
@@ -0,0 +1,49 @@
...
+const monthNames = ['January', 'February', 'March', 'April', 'May', 'June', 'July', 'August', 'September', 'October', 'November', 'December'];
+
+magic(readme, {
+  transforms: {
+    TC39(content, options) {
+      return dedent(byYear.map(([year, months]) => {
+        return `<details${isOpen(year)}>
+          <summary>${year}</summary>
+
+            ${dedent(months.map(m => `- [${monthNames[m - 1]}](./${year}/${m}.md)`).join('\n'))}
```
michaelficarra: Reduce this indentation one level to fix generated month list indentation.

scripts/package.json
ljharb: It’s just `./scripts/generate.js` right now - the reason it’s this way is so the root of the repo doesn’t get cluttered with package.json and .npmrc. If that’s not an issue, i can move them all up, and then `npm run build` would work?

scripts/generate.js
ljharb: thanks, fixed

scripts/package.json
ljharb: Updated the OP with more explanation of this.

PR #300, opened 2018-01-16
Add agenda item for optional-chaining updates.

gisenberg:


littledan:
Should this be in the late agenda items bucket, or is that only for things that propose stage advancement?

ljharb:
whoops, closed this by mistake

ljharb:
@littledan That's only for things that propose stage advancement.

ljharb:
Merged in cd101f49e9615ec39065c977f557e76f75ffdeac

PR #323, opened 2018-03-01
Add Module Keys Stage 0 Strawman to agenda

mikesamuel:
Per [Entrance Criteria](https://tc39.github.io/process-document/) for Stage 1:

* [X] Identified “champion” who will advance the addition - Me
* [X] Prose outlining the problem or need and the general shape of a solution - See linked
* [X] Illustrative examples of usage - See linked
* [X] High-level API - See linked
* [X] Discussion of key algorithms, abstractions and semantics - See linked
* [ ] Identification of potential “cross-cutting” concerns and implementation challenges/complexity - Will pursue between now and May

PR #324, opened 2018-03-02
Request a 30 minute slot for Module Keys proposal up from 15 minutes

mikesamuel:
Per discussion with @littledan in https://github.com/tc39/agendas/pull/323

### DIFF
--- a/2018/05.md
+++ b/2018/05.md
@@ -87,8 +87,8 @@ Supporting materials includes slides, a link to the proposal repository, a link 
 1. Timeboxed agenda items
     1. 15-minute items
         1. [Sequence properties in Unicode property escapes](https://github.com/mathiasbynens/proposal-regexp-unicode-sequence-properties) for Stage 1 (Mathias Bynens)
-        1. [Module Keys](https://github.com/mikesamuel/tc39-module-keys) for Stage 1 (Mike Samuel)
     1. 30-minute items
+        1. [Module Keys](https://github.com/mikesamuel/tc39-module-keys) for Stage 1 (Mike Samuel)
     1. 45-minute items
     1. 60-minute items
 1. Non-timeboxed overflow from previous meeting


michaelficarra:
@mikesamuel Given the potential breadth of this topic, I'd recommend a still longer timebox, or even no timebox at all.

mikesamuel:
@michaelficarra I'll put together some slides and time myself.  Do you have a sense for what might be the most controversial parts?

littledan:
I'm not sure if a longer timebox is needed. It's true that it could be difficult to deeply discuss everything completely in 30 minutes, or even an hour, but sometimes the first discussion about a topic is more an introduction, to motivate research and discussion offline. A non-timeboxed presentation risks falling off the bottom of the agenda and also engaging only the small percentage of the committee members who take the time to think deeply about the proposal before the meeting.

leobalter:
I strongly believe 30 minutes won't be enough for this topic, as I believe it's at least worth having a non superficial discussion over this topic if this will ever get to advance in the stage process.

Although, the current timebox should be enough to present and start some discussions, even if not enough to capture all the questions and concerns.

leobalter:
@mikesamuel are you sure you want to add this for the meeting in May and not the next in March?

I just want to confirm this to avoid any frustration.

In the meanwhile, @dherman and @bmeck you might be interested in this topic.

mikesamuel:
@leobalter I'm excited to talk about this, but I won't be able to make it to London in March.  At best I could do a call-in, but I am not good at relating unfamiliar concepts over the phone.   I live one state over from New York state, so can easily make it to the May meeting.

leobalter:
sure thing, I'm looking forward to discuss this in May.

littledan:
@mikesamuel A call might not be a bad idea anyway, e.g., to briefly introduce the concept. I've introduced proposals and advanced via a call before (e.g. private methods went straight to Stage 2 while I was on the other side of the pond) so I think it can be workable.

PR #325, opened 2018-03-05
Add "An evening with TC39 in London" event

keithamus:


IgnoredAmbience:
Although, it's getting to the point where I'm considering moving the wordier bits of the logistics section elsewhere, to prevent cluttering the agenda too much.

keithamus:
Happy to follow your lead there @IgnoredAmbience 😃

littledan:
Merging the patch to add this to the agenda; any moves can take place in a follow-on patch.

PR #341, opened 2018-03-16
Add Local Information for March Meeting

IgnoredAmbience:
And split out excess detail to new document.

(Pull request is to ensure people get notifications of this large, quite late change)

### DIFF
--- a/2018/03.md
+++ b/2018/03.md
@@ -8,12 +8,26 @@
   - 10:00 to 17:00 UTC on 20 March 2018
   - 10:00 to 17:00 UTC on 21 March 2018
   - 10:00 to 16:00 UTC on 22 March 2018
-- **Location**: Imperial College London, South Kensington Campus, Exhibition Road, London, SW7 2AZ, United Kingdom
-  - [Map](https://goo.gl/maps/jMhXKeuVXFm)
-  - [Travel Directions](http://www.imperial.ac.uk/visit/campuses/south-kensington/)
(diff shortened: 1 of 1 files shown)


littledan:
Will the dinner be sponsored by ICL and Ecma, or will it be paid by the attendees individually?

IgnoredAmbience:
We've tried to contact Ecma to verify whether this arrangement still stands but didn't receive a response in time to confirm this. At present, the plan is for ICL to make a contribution towards drinks and some snack platters as appetisers, unfortunately, our budget doesn't allow for funding the whole dinner for this number of delegates.

littledan:
@IgnoredAmbience Thanks for arranging for the catering; I really appreciate it. This helps create a social space for TC39 members to talk in and around the meeting.

> We've tried to contact Ecma to verify whether this arrangement still stands but didn't receive a response in time to confirm this.

Which arrangement are you referring to?

What I've heard on the Ecma side so far is that if you can't provide funding, then attendees will pay for their own meals and drinks on Wednesday night. If this is the case, I think the agenda should be clear about it as soon as possible so that attendees can make their plans accordingly.

IgnoredAmbience:
>> We've tried to contact Ecma to verify whether this arrangement still stands but didn't receive a response in time to confirm this.
> Which arrangement are you referring to?

Whether Ecma can share costs, I believe we asked for reasonable budget limits when planning for the venue. I'll need to double check with Teresa precisely what we asked.

IgnoredAmbience:
I've added a note to this effect to the agenda.

littledan:
cc @ecmageneva @RexJaeschke Could you clarify whether Ecma plans to share the cost for the dinner?

PR #345, opened 2018-03-18
Await in parameter defaults for non-arrow async functions

pemrouz:
/cc @littledan


littledan:
Although this is a late addition, Pedram has been working on this topic for months and travel to the US is difficult. I hope you all don't mind adding this agenda item a bit late.

PR #354, opened 2018-04-18
Add `export * as ns from "mod”` PR

leobalter:
https://github.com/tc39/ecma262/pull/1174


ljharb:
Is https://github.com/tc39/test262/pull/1498 otherwise ready to land? Mergeable tests were a precondition, as i recall, for coming back to the committee.

leobalter:
@ljharb tests are ready to land, yes, but we might end up with another review. Sorry for pressing merge before getting your question.

PR #359, opened 2018-04-28
approving a replacement name for Array.prototype.flatten (not smoosh)

michaelficarra:


mathiasbynens:
So looking forward to this.

jdalton:
So rad if folks tackled the root cause instead of bikeshedding names.

littledan:
I'd suggest that you start the discussion of this name with TC39 before the meeting somehow so we can get further in the committee discussion.

michaelficarra:
@littledan https://github.com/tc39/Reflector/issues/142

PR #373, opened 2018-05-15
Add how we work project and website update

codehag:


codehag:
I dont think we will give a super substantial update, only showing where the website is at, and the current status of the docs, and give a call to action to people present on specific topics

codehag:
Ah I missed that, I thought it was the 15 min. I will change it

PR #379, opened 2018-05-21
Add SuperProperty evaluation order presentation

jridgewell:
This is super late, so I'm not sure if it should be accepted.

If it is, it should be the lowest priority item. Maybe a time-filler if we can't get all the longer presentations items in?

And if it's not accepted, I'll do better getting it onto the next meeting's agenda on time.


littledan:
I'm pretty sure we're going to run over and not complete the agenda. Adding this as a low-priority item, or just delaying it until next meeting, sounds good to me.

mathiasbynens:
This is not a proposal seeking stage advancement, so per https://github.com/tc39/agendas/blob/master/2018/05.md#agenda-topic-rules there is no deadline.

jridgewell:
> Adding this as a low-priority item

Is there a way to mark it low priority? I figured it would just be telling Rex at the start of the meeting.

> so per [agenda-topic-rules](https://github.com/tc39/agendas/blob/master/2018/05.md#agenda-topic-rules) there is no deadline

I'd still rather have given the browsers time to look over it. I'm fine with whichever decision we make.

PR #387, opened 2018-05-23
Add overflow topic to revisit Function.prototype.toString pragma

littledan:


2018/05.md
```diff
@@ -177,6 +177,7 @@ Supporting materials includes slides, a link to the proposal repository, a link
 
     | topic | presenter |
     |-------|-----------|
+    | [Function.prototype.toString() censorship](https://domenic.github.io/proposal-function-prototype-tostring-censorship/) for stage 2 | Domenic Denicola |
```
michaelficarra: Can we change the topic? This reached stage 2 yesterday.

ljharb: Consensus is currently listed as "Stage 2, but there seems to be objections?" so i think its status is still in doubt

ljharb: I'm going to merge to get it on the agenda; but feel free to discuss a topic change as a followup

domenic: I think the right way to proceed is to take yesterday's lack of concrete objections at face value, and if some concrete objections materialize in this overflow session, then move it backwards. I look forward to hearing such objections and how they could be addressed in plenary, if they have become more concrete than yesterday.

PR #392, opened 2018-05-23
Add direct link to slides, adjust timebox

rbuckton:


rbuckton:
I shortened the timebox for two Stage 1 proposals (though we may not even get to them).

Issue #397, opened 2018-06-06
Scheduling May 2018's overflow agenda items


littledan:
In the May 2018 TC39 meeting, we decided to prioritize later stage proposals and short discussions over longer discussions and earlier stage proposals. I still think this was a good idea, but now we have a bit of a backlog of proposal introductions and discussion topics from last meeting, which may be good to get back to in the July 2018 meeting.

I want to suggest a way to schedule these topics in a way which might help us get through everything while not pushing out the later stage proposals that we also need to discuss:
- For Stage 0/1 proposals, let's discuss them in a 10-minute lightning talk sessions, which are oriented towards presenting a proposal to the committee and seeking feedback on GitHub (suggestion from @kosamari ). We could reserve Thursday morning for these lightning talks.
- For discussion topics, let's discuss these Thursday afternoon. By the end of TC39 meetings, we're all a bit tired, so let's use this time to reflect and think at a broader level what we'd like to do next.

Thoughts?

domenic:
I'm unsure what you're suggesting, but one interpretation of it is that my topics got pushed back from one meeting, and now are getting restricted to 10-minute lightning talks in the next meeting. That would mean I would need three meetings before getting the longer discussion I originally scheduled, which would be very frustrating.

Is that the proposal?

michaelficarra:
@domenic Do you not agree with the strategy of prioritising later stage proposals over newer ones to reduce our number of active proposals? The fact that newer ones are getting pushed off the end of the agenda shows that we have too many in the pipeline given available resources (3 days/2 months). It will only be like this until we reach a manageable number of active proposals, at which point it reaches a natural equilibrium.

domenic:
Yeah, I do not agree with that strategy; I think some consideration has to be given to people who want to present their own work, even if other peoples' older proposals have had a hard time advancing. It's a balance, certainly.

michaelficarra:
I understand and empathise with those who have earlier stage proposals at this time. I think that the balance you're requesting is what Dan is suggesting. That will allow the champion to at least publicise their idea and solicit committee feedback through GitHub. Do you have an alternative idea?

domenic:
I preferred the balance where your proposal already got deprioritized (regardless of how early you had it ready for the agenda) by placing it below higher-stage proposals at one meeting. But you did not get continually punished for all future meetings, and instead we allowed overflow items to go first as they always have.

michaelficarra:
[...]

PR #400, opened 2018-06-20
Reorder discussion topics, add lightning talks

littledan:
Reordering per https://github.com/tc39/agendas/issues/397#issuecomment-398423845
- Add a "Stage 0 lightning talks" section, 10 minutes per topic
- Put discussion topics after Stage 0 lightning talks, and remaining new Stage 0 proposals after that
- Maintain the placement of overflow items at the front; we can consider changes to handle overflow buildup in a follow-on PR

Let's make sure to get feedback from the people who voiced opinions on #397 before merging. cc @leobalter @ljharb @michaelficarra @domenic

### DIFF
--- a/2018/07.md
+++ b/2018/07.md
@@ -72,7 +72,7 @@ Supporting materials includes slides, a link to the proposal repository, a link 
     | timebox | topic | presenter |
     |:-------:|-------|-----------|
 
-1. Timeboxed overflow from previous meeting
+1. Overflow from previous meeting
 
     | timebox | topic | presenter |
     |:-------:|-------|-----------|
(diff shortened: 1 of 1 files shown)


littledan:
Thanks to @kosamari for the idea of adding lightning talks.

2018/07.md
```diff
@@ -72,7 +72,7 @@ Supporting materials includes slides, a link to the proposal repository, a link
     | timebox | topic | presenter |
     |:-------:|-------|-----------|
 
-1. Timeboxed overflow from previous meeting
```
ljharb: What about non-timeboxed overflow? We previously had two sections, so that timeboxed overflow would retain priority over non-timeboxed overflow. I see the "(overflow)" below; should we use an emoji (like the ⌛️ for late items) instead?

littledan: I didn't see the non-timeboxed overflow section. I'd like to propose that we just put all of the overflow topics in one section, sorted in the order that they were sorted in the previous meeting's agenda, but not broken out into sub-categories.

littledan: If you have a proposal for an overflow emoji, happy to switch to it. Looks like a nice way to present the information.

ljharb: Perhaos ➿ or ➰ or 🔁 or ♻️ or 🙉 or 🛅 or ⏪ ?

What do you see as the advantage of having all overflow items in one section? Would all the non-timeboxed ones have an empty "timebox" column?

2018/07.md
```diff
@@ -72,7 +72,7 @@ Supporting materials includes slides, a link to the proposal repository, a link
...
+1. Overflow from previous meeting
 
     | timebox | topic | presenter |
     |:-------:|-------|-----------|
```
domenic: Hmm I'm not sure what to do with blooks which were scheduled for 45 minutes but got 10 minutes.

I guess I'm OK in this particular instance not reserving time, but I think we should be cautious in the future of assuming that giving someone a ten-minute lightning talk is enough to remove their overflow privileges.

2018/07.md
littledan: The advantage is, now that we have several different boxes, it would get unwieldy to have so many different overflow sections. It's just extra syntactic overhead that doesn't serve us, since people will not be adding their own new topics and figuring out where they go.

2018/07.md
littledan: I don't think overflow "privileges" are something we should be relying on generally. The idea is to try to fit everyone's topic in, and we all compromise on time to make that happen, using overflow as a last resort.

PR #412, opened 2018-07-12
Include Dynamic Modules Proposal

guybedford:
Will provide an update soon with the presenter once I've got this confirmed.

### DIFF
--- a/2018/07.md
+++ b/2018/07.md
@@ -106,7 +106,6 @@ Supporting materials includes slides, a link to the proposal repository, a link 
     | 1     | 30m     | [RegExp match offsets](https://github.com/tc39/proposal-regexp-match-offsets) for Stage 2 | Ron Buckton |
     | 1     | 60m     | [Partial Application](https://github.com/tc39/proposal-partial-application) Update | Ron Buckton |
     | 1     | 60m     | [Cancellation](https://github.com/tc39/proposal-cancellation) Update | Ron Buckton |
-    | 0     | 60m     | [Dynamic Modules Proposal](https://github.com/guybedford/proposal-dynamic-modules) for Stage 1 | TBD |
 
 1. Stage 0 10m lightning talks (Thursday morning)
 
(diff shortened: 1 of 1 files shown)


2018/07.md
```diff
@@ -106,6 +106,7 @@ Supporting materials includes slides, a link to the proposal repository, a link
     | 1     | 30m     | [RegExp match offsets](https://github.com/tc39/proposal-regexp-match-offsets) for Stage 2 | Ron Buckton |
     | 1     | 60m     | [Partial Application](https://github.com/tc39/proposal-partial-application) Update | Ron Buckton |
     | 1     | 60m     | [Cancellation](https://github.com/tc39/proposal-cancellation) Update | Ron Buckton |
+    | 0     | 60m     | [Dynamic Modules Proposal](https://github.com/guybedford/proposal-dynamic-modules) for Stage 1 | TBD |
```
ljharb: i think for this meeting we were trying for stage 0 proposals to have 10 minute lightning talks (see the next section). do you anticipate this taking an hour to determine "the committee expects to devote time to examining the problem space, solutions and cross-cutting concerns"?

littledan: BTW you can make a longer than 10 minute talk for a Stage 0 proposal, it'll just be prioritized later on the agenda.

guybedford:
Thanks, I've moved this to the appropriate Stage 0 section.

littledan:
The current section looks good to me. From communication with @guybedford , it seems that there are several possible TC39 delegate presenters. I'm merging this patch now, so that we have it on the agenda ahead of the deadline, and we'll figure out who the presenter is in the coming days.

ljharb:
I think it should have waited until there’s a named TC39 member associated with it - who are the possible presenters?

littledan:
We are approaching the deadline, so I think the high order bit is to make the topic visible to other attendees. The practical impact is to give people a chance to review the proposal; the name of the TC39 champion doesn't affect much for preparation.

When it comes to commitments like presenting at a meeting, it can be awkward to name yourself publicly as a possible speaker before you are ready to commit. Let's give @guybedford a chance to work this out with the potential presenters before trying to force a commitment.

PR #416, opened 2018-07-15
Fix date of future meeting

MasterCassim:
Fixed the date of the meeting at GoDaddy. Date was specified as 2018.


littledan:
Thanks for the fix!

PR #423, opened 2018-07-23
Remove day/time suggestions

littledan:
Thursday morning/afternoon for certain topics seems unrealistic; let's work out the details later.


codehag:
@ljharb I am not available in the afternoon

ljharb:
Obviously we have to work around schedule constraints, but confining lightning talks to Thursday was part of reformatting the agenda in #400.

littledan:
We don't have anyone signed up for lightning talks at this point, so it seems pretty unrealistic to allocate a morning to it.

ljharb:
I don't read the heading as allocating the morning to it, more saying that if any, that's when they'd be discussed.

littledan:
I think the whole agenda is clearer if we remove these times; that's why the PR does it.

PR #425, opened 2018-07-23
Decorators not going to Stage 3 just yet

littledan:


littledan:
@michaelficarra If you're cheering this, I'd be happy to have your feedback on the proposal and any issues you have with it.

PR #433, opened 2018-07-25
Add slides for "await ticks" normative PR

zenparsing:


leobalter:
@ljharb merge this and we follow up with a link in the agenda? I can do it if you're thumbs up

PR #436, opened 2018-09-06
re-order proposal topics according to stated order

michaelficarra:


ljharb:
I’m confused, which stated order?

ljharb:
ah, nvm, i see it.

PR #455, opened 2018-09-20
Add agenda item for JSON.parse source text access

gibson042:


gibson042:
Understood. This is a "time permitting" proposal for addressing issues I've seen come up recently with BigInt and Temporal.

ljharb:
Talking about it is fine; but being past the deadline means that it can't advance at this meeting. I'll tweak the entry and merge.

michaelficarra:
@ljharb To be pedantic, the committee did not limit its power with this rule. Instead, we granted individual representatives explicit permission to reject advancement for any late proposals solely for the reason that they were submitted late. Previously there was criticism of this practice. If every present representative is in favour of a late proposal advancing, it may still be advanced.

ljharb:
You’re correct, that is a very good point - i feel like every time it’s come up, there’s been pushback, so maybe i have a more pessimistic view of it.

michaelficarra:
We should probably update the wording on the agenda to better reflect the actual agreed upon rule for late agenda items.

PR #470, opened 2018-11-05
WebAssembly modules update

littledan:
Should this be handled all on GitHub instead? Editors, if you decide to land this patch before the meeting, no need to discuss there.


littledan:
Merging the PR; let's remove this agenda item if the editors decide to merge the patch earlier.

PR #472, opened 2018-11-12
Update Agenda

dustinsavery:
After discussing with @littledan, I'm going to hold off on asking for Stage 2 for now and just give an update.


2018/11.md
```diff
@@ -108,6 +108,7 @@ Supporting materials includes slides, a link to the proposal repository, a link
     | 5m      | Override mistake fix status update ([PR](https://github.com/tc39/ecma262/pull/1320)) | Daniel Ehrenberg |
     | 1m      | Meeting planning follow-up ([dates/locations](#dates-and-locations-of-future-meetings); [Issue](https://github.com/tc39/Reflector/issues/130)) | Daniel Ehrenberg |
     | 15m     | `Array.prototype.sort` stability ([slides](https://docs.google.com/presentation/d/1Io53b2Bi3_N0_wguWoA9OKuPRpHch34EVbS1H8zISes/edit), [PR](https://github.com/tc39/ecma262/pull/1340)) | Mathias Bynens |
+    | 30m     | Update on Optional Chaining [Optional Chaining](https://github.com/tc39/proposal-optional-chaining) | Dustin Savery |
```
ljharb: I think this should maybe still be in the proposals section; that’s not just for stage advancement.

dustinsavery: Ahh, I was not aware of that. In that case, I will just update the label.

PR #483, opened 2018-11-16
Sort proposals in November agenda by stage

chicoxyzzy:
Hashbang grammar proposal is breaking the order


michaelficarra:
Actually should've gone in front of the other stage 2 proposals since we sort both by stage (descending) and timebox (ascending).

PR #497, opened 2019-01-07
add Security Implications of Error.prototype.stack discussion

michaelficarra:


michaelficarra:
It's just a simple example of information leakage through call stack inspection, and what I plan to do about it through the committee.

ljharb:
I’d love to chat about it with you as it relates to the stacks proposal; I’ll ping you.

PR #514, opened 2019-01-18
fix @ginden's name

michaelficarra:
@Ginden is it true that you will be presenting this topic with @gsathya at the upcoming meeting?


littledan:
I wouldn't mind acknowledging proposal authors in the agenda, but I'm not sure the right way to present this information.

PR #515, opened 2019-01-20
Built-in decorators proposal

littledan:


littledan:
Does this mean it should be limited to a Stage 0 presentation, or what?

ljharb:
It means that anyone may object to advancement solely on the basis of missing the deadline; it certainly still could advance.

PR #521, opened 2019-01-22
Overloading method parameters between BigInt and Number: Just Say No

littledan:
Accidentally added this in a commit without a PR


ljharb:
https://github.com/tc39/agendas/commit/85c4199dafbb56773b492425058116e12f04e1d3

PR #533, opened 2019-01-29
agenda: built in specifiers

MylesBorins:
I was planning to bring this up during discussion around
"Standard Library" which appears to have been removed from the agenda.
Is it too late to add this? Requesting 30 minutes but would settle for
even 5 minutes

Refs: https://github.com/tc39/proposal-javascript-standard-library/issues/12
Refs: https://github.com/nodejs/node/pull/21551


MylesBorins:
FWIW, what is really important for me here is to have some sort of discussion on the record that we can reference in node for future decision making

codehag:
we should have time

PR #559, opened 2019-03-25
Decorators slides

littledan:


jridgewell:
@littledan Is [slide 28](https://docs.google.com/presentation/d/1mGqWHfs1EkBneG9CGJZj5mefzMrXQnt2FjqRp9geGxI/edit#slide=id.g50c437bc8b_0_94) correct? Seems like it's a copy paste of the `@wrap` desugaring.

rwaldron:
Slide 26, right hand code pane: `f.call(this, ...args);` should be `f.apply(this, ...args);` (I would've left a comment on the slide, but I'm mobile and don't see any comment option)

nicolo-ribaudo:
@rwaldron It is correct with `call`; if you use `apply` you should remove `...`.

littledan:
@jridgewell Thanks for pointing out this error. I've fixed it.

(And, if you tried to access the document initially, I forgot to fix the sharing settings; it's now public.)

bathos-wistia:
Apologies if this is too off-topic, but are there any resources published yet which explain the `.decorators.json` slides? I was able to follow up until that point. (And it looks pretty great!)

nicolo-ribaudo:
Only https://hackmd.io/44ErLPn8Qi6FshyoTcrXcA, which is how we plan to implement static decorators in Babel.

PR #577, opened 2019-05-24
Add ECMA-402 Normative PRs

leobalter:


littledan:
These are fundamentally new features, not bug fixes. I wonder if we should prioritize them as such, with the staged proposals.

ljharb:
Adding options definitely seems like proposal territory.

littledan:
Well, maybe "stages" as such would be a little overkill process-wise, but I think we can follow the same spirit (and priorities) for PRs.

leobalter:
The list is not exclusive to bugs, and includes "Needs Consensus PRs".

I don't mind where they are prioritized in the agenda if they get addressed during the meeting. Perhaps, the actual matter is asking @FrankYFTang of time constraints as he might be attending the meeting from a far timezone (PST), I'll be on the EST and my preference is to only avoid anything in the mornings.

@littledan should we move these items or just let the chairs know they are not bugs?

littledan:
Deferring to the chairs on agenda ordering SGTM. @FrankYFTang , if you have any time constraints, please write them on the agenda in a PR.

ljharb:
It’s not about bugs, really - the category is meant for non-proposal things that can be landed with consensus. It’s certainly reasonable to try to land these as needs consensus PRs (but to me they do seem like they should be proposals, since they’re adding new features)

leobalter:
as some people discussed @ Intl, these are just new options for the existing API, not adding new programming logic. They are pretty straight forward.

My suggestion for these will be this simple pattern:

1. Discuss and achieve consensus for the new featured options.
2. Get at least 1 implementation shipping without flags.
3. Get the other implementors agreeing to eventually implement this.
4. Implementation pass on landing test262 tests.
5. If it proves being web incompatible or any other problems are found, return to 1.
6. Land proposed text to the specs.

I believe this pattern is used often in other PRs, and believe it's fair enough here. WDYT, @ljharb?

ljharb:
Sure, i think it’s a reasonable path! I have a loose preference for “new options” to be considered just as logic-full and proposal-y as other new features, but that’s for the plenary to decide.

PR #597, opened 2019-07-11
Add map updateOrInsert to July 2019 Agenda

bmeck:


domenic:
Slides are currently private

bmeck:
woops, fixed

zloirock:
Looks like an equal of `Map.prototype.update` from [this proposal](https://github.com/tc39/proposal-collection-methods).

bmeck:
@zloirock that one doesn't switch on if the element is in the map or not

zloirock:
So maybe makes sense to add it in the scope of the mentioned proposal?

bmeck:
@zloirock I don't want to hold up this method on all the other methods in the proposal you mentioned.

michaelficarra:
@bmeck I think it would help the proposal if you include usage/popularity data from libraries that shows this is a very common need, especially given how easy it is for one to implement on their own.

bmeck:
@michaelficarra I can do some code searches, they have been more telling than library audits. It is simple to do these patterns so much that few libraries even in other languages seem to provide too much. People simply write if statements a lot. Usage % I have no clear way of measuring

PR #602, opened 2019-07-20
Add casing to open-ended discussions

sffc:
@leobalter @ljharb I made some slides.  I think it is worth discussing this topic with TC39.  Thoughts?


leobalter:
I’m glad you’re doing this, thank you, @sffc!

PR #605, opened 2019-07-23
remove backslashes from Symbol.matchAll agenda item

michaelficarra:


ljharb:
this breaks the link :-/

PR #609, opened 2019-07-24
Michael Fig new scheduling constraint

erights:


leobalter:
While I don't control the scheduling constraints we need to be more vigilant to our time constraints as this change might cause a lot of stress for those organizing the meetings. It's a bit harsh to have this information after the second day of the meeting has already started.

I hope delegates find a better organization to support better how the topics will flow.

gesa:
It is p rough but @erights did talk to me before adding this

leobalter:
thanks for the clarification, @gesa

erights:
> I hope delegates find a better organization to support better how the topics will flow.

I agree. I will try to do better in the future. Thanks for accommodating this time.

PR #636, opened 2019-10-02
replace records and tuples slides link with permanent version

rickbutton:
The new link contains exactly the same slide deck bit-for-bit, but I've hosted it in a permanent location instead of a temporary dropbox.


rickbutton:
@michaelficarra good call, done

PR #677, opened 2020-01-24
Add duration format proposal for stage 1

younies:


younies:
@jridgewell  ... Thanks for the note. Done :)

PR #682, opened 2020-01-29
process changes to accommodate US members and US delegates

michaelficarra:
I would like this discussion to pre-empt any technical topics, as proposed changes (if accepted) would apply during this meeting. I may be able to get a few slides later this week or this weekend.

/cc @tc39/chair-group If you're available for a call before the meeting, I can brief you on the proposal and possibly align with any plans you already have. @MylesBorins is already pretty familiar with the proposal.


littledan:
Are there any details you could publish to the committee (either publicly or in the Reflector) ahead of the meeting? It would certainly help Igalia if we could review the proposal ahead of time.

PR #708, opened 2020-03-17
New agenda item:  'OptionalChain'.PrivateIdentifier as valid syntax

caiolima:
...


rkirsling:
Slide content note: looks like it's tc39/proposal-class-fields#301, not 30.

caiolima:
> Slide content note: looks like it's [tc39/proposal-class-fields#301](https://github.com/tc39/proposal-class-fields/pull/301), not 30.

Nice catch! Fixed.

PR #710, opened 2020-03-19
Ask for 30 min timebox on agenda item

caiolima:
I'd like to ask for 30 min, if possible because I think it is very likely that discussion of this PR can last more than 15 min based on this [thread](https://github.com/tc39/proposal-class-fields/issues/302).


jridgewell:
We will probably need to update the slides, too.

PR #716, opened 2020-03-23
Add Remote Plenaries agenda item

robpalme:
This is to discuss the format of Remote Plenaries going forwards.


robpalme:
Probably this ought to be scheduled at the start of the third-day, once we all have some feel of how the first two days have gone.

PR #724, opened 2020-03-26
[WIP] update agenda template; fixes #722

### LINKED ISSUE #722: update requirements for stage 0 and stage 1 on the agenda document
We have been finding it difficult to review proposals that have no materials. Sometimes this happens with stage 1 proposals, as there is no strict requirement that they have materials present before the 10 day deadline. It is pretty rare that a stage 1 proposal does not advance, so having materials present isn't as crucial as for later stages. However, if all we have to go by is a title, it is difficult to prepare for the meeting in advance, or have preliminary discussions with champions. This in the end results in a less productive discussion during plenary.

How do people feel about changing the stage 1 requirement in the agenda template to having a ready repo with the proposal written out in prose, and a clear problem statement? This can be generated from the proposal template. We can help newer members adjust though mentoring.

codehag:
for discussion next week

### DIFF
--- a/template.md
+++ b/template.md
@@ -20,8 +20,10 @@ Deadline for advancement eligibility: [**CALCULATE TIME HERE**](https://www.time
   - <sub>Note: this time is selected to be precisely 10 days prior to the start of the meeting</sub>
 
 1. Proposals not looking to advance may be added at any time; if after the deadline, please always use a pull request so that members are notified of changes.
-1. Proposals looking to advance to stage 0 or 1 must be added (and noted as such) prior to the deadline, or else delegates may object to advancement solely on the basis of missing the deadline.
+1. Proposals seeking feedback at stage 0 must be added (and noted as such) prior to the deadline, or else delegates may object to advancement solely on the basis of missing the deadline.
     1. Such proposals *should* include supporting materials when possible.
+1. Proposals looking to advance to stage 1 must be added (and noted as such) prior to the deadline, or else delegates may object to advancement solely on the basis of missing the deadline.
+    1. Such proposals *must* link to a proposal repository and they *should* link to supporting materials when possible.
 1. Proposals looking to advance to stages 2, 3, or 4 must be added (and noted as such) *along with the necessary materials* prior to the deadline, or else delegates may object to advancement solely on the basis of missing the deadline.
     1. Such proposals *must* link to supporting materials prior to the deadline, or else delegates may object to advancement solely on the basis of missing the deadline. If these materials change substantially after the deadline, the proposal may be disqualified from advancement, based on the committee’s judgment.
     1. Proposals looking to advance to stage 4 *must* link to a pull request into [the spec](https://github.com/tc39/ecma262), since the [process](https://tc39.github.io/process-document/) requires one.


template.md
```diff
@@ -20,8 +20,10 @@ Deadline for advancement eligibility: [**CALCULATE TIME HERE**](https://www.time
   - <sub>Note: this time is selected to be precisely 10 days prior to the start of the meeting</sub>
 
 1. Proposals not looking to advance may be added at any time; if after the deadline, please always use a pull request so that members are notified of changes.
-1. Proposals looking to advance to stage 0 or 1 must be added (and noted as such) prior to the deadline, or else delegates may object to advancement solely on the basis of missing the deadline.
+1. Proposals seeking feedback at stage 0 must be added (and noted as such) prior to the deadline, or else delegates may object to advancement solely on the basis of missing the deadline.
     1. Such proposals *should* include supporting materials when possible.
+1. Proposals looking to advance to stage 1 must be added (and noted as such) prior to the deadline, or else delegates may object to advancement solely on the basis of missing the deadline.
+    1. Such proposals *must* include a proposal repository and they *should* include supporting materials when possible.
```
jridgewell:
```suggestion
    1. Such proposals *must* link to a proposal repository and they *should* link to supporting materials when possible.
```

template.md
```diff
@@ -20,8 +20,10 @@ Deadline for advancement eligibility: [**CALCULATE TIME HERE**](https://www.time
   - <sub>Note: this time is selected to be precisely 10 days prior to the start of the meeting</sub>
 
 1. Proposals not looking to advance may be added at any time; if after the deadline, please always use a pull request so that members are notified of changes.
-1. Proposals looking to advance to stage 0 or 1 must be added (and noted as such) prior to the deadline, or else delegates may object to advancement solely on the basis of missing the deadline.
+1. Proposals seeking feedback at stage 0 must be added (and noted as such) prior to the deadline, or else delegates may object to advancement solely on the basis of missing the deadline.
```
jridgewell: If these are just seeking feedback, does the "object to advancement" statement apply to them?

codehag: I am not sure what to do with stage 0, we can't object in those cases but im not sure what to do here.

jridgewell: Me neither.

rkirsling: Something like this, perhaps?
```suggestion
1. Proposals seeking feedback at stage 0 must be added (and noted as such) prior to the deadline, or else delegates may object to the presentation solely on the basis of missing the deadline.
```

template.md
```diff
@@ -20,8 +20,10 @@ Deadline for advancement eligibility: [**CALCULATE TIME HERE**](https://www.time
...
     1. Such proposals *should* include supporting materials when possible.
+1. Proposals looking to advance to stage 1 must be added (and noted as such) prior to the deadline, or else delegates may object to advancement solely on the basis of missing the deadline.
+    1. Such proposals *must* link to a proposal repository and they *should* link to supporting materials when possible.
```
rkirsling: Judging from what follows, the proposal repository _is_ an example of necessary supporting materials—perhaps the stage 1 case should just be merged into the stage 2/3/4 case below?

template.md
ljharb: I don't think anyone can (or should be able to) object to a presentation, generally.

Typically, we've considered something stage 0 when it's either been presented to the committee, or is on the agenda to be, and the only reason something won't qualify for stage 0 (or will be evicted from it) is if there's zero likelihood it will proceed.

ljharb:
This received consensus today.

PR #746, opened 2020-05-01
Update schedule for 4 day 5 hour sessions.

MylesBorins:
Based on prior conversation

Thoughts?

### DIFF
--- a/2020/06.md
+++ b/2020/06.md
@@ -6,9 +6,10 @@
 - **Host**: Paypal
 - **Dates**: June 2 - 4, 2020
 - **Times**:
-  - 10:00 to 17:00 CT on June 2nd, 2020
-  - 10:00 to 17:00 CT on June 3rd, 2020
-  - 10:00 to 16:00 CT on June 4th, 2020
+  - 10:00 to 15:00 CT on June 1st, 2020
+  - 10:00 to 15:00 CT on June 2nd, 2020
+  - 10:00 to 15:00 CT CT on June 3rd, 2020
+  - 10:00 to 15:00 CT CT on June 4th, 2020
 - **Location**: Remote
 - **Attendee information**: https://github.com/tc39/Reflector/issues/279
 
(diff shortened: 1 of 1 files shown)


2020/06.md
```diff
@@ -6,9 +6,10 @@
 - **Host**: Paypal
 - **Dates**: June 2 - 4, 2020
 - **Times**:
-  - 10:00 to 17:00 CT on June 2nd, 2020
-  - 10:00 to 17:00 CT on June 3rd, 2020
-  - 10:00 to 16:00 CT on June 4th, 2020
+  - 10:00 to 15:00 CT on June 1st, 2020
+  - 10:00 to 15:00 CT on June 2nd, 2020
+  - 10:00 to 15:00 CT CT on June 3rd, 2020
+  - 10:00 to 15:00 CT CT on June 4th, 2020
```
nicolo-ribaudo, line 12:
```suggestion
  - 10:00 to 15:00 CT on June 3rd, 2020
  - 10:00 to 15:00 CT on June 4th, 2020
```

littledan: Should we use America/Chicago to be more unambiguous? :)

Jack-Works:
Look not good to me. 10:00 to 15:00 CT is 11 PM to 4 AM for me. Stay up for 4 days is too hard.

littledan:
These aren't during my workday either, but for me, it's easier to stay up for shorter amounts of time over more days, than longer amounts of time for fewer days. The four-day schedule was already a compromise, so I'm +1 on this direction.

MylesBorins:
This is what we had agreed to in the last meeting, specifically to stick to the times that were local to the area the meeting was planning to be held.

I'm going to land this and open up an issue to discuss alternatives. It is worth mentioning that after the last meeting it was left that the chair team should be a final arbitrator of this... so want to clarify that discussion is not for the purpose of building consensus but rather to look for ideas that we may be able to adopt to make the schedule for flexible for folks

PR #756, opened 2020-05-22
add sublcassing removal proposal

codehag:
This is a placeholder for now, waiting on Shu's ok for adding this...


codehag:
shu said it was ok

PR #815, opened 2020-07-10
Add cleanupSome to weakrefs

codehag:


littledan:
I guess the plan is to propose it for Stage 3, right? Maybe we don't need to mention "2" in the agenda item.

PR #824, opened 2020-07-14
Add Strictness check for obj's SetMutableBinding to 2020/07

leobalter:


2020/07.md
```diff
@@ -61,6 +61,7 @@ Supporting materials includes slides, a link to the proposal repository, a link
     |   | 15m     | Should eval?.() be direct eval? ([#2062](https://github.com/tc39/ecma262/issues/2062), [#2063](https://github.com/tc39/ecma262/pull/2063)) | Ross Kirsling |
     |   | 15m     | [Host hooks for Job callbacks](https://github.com/tc39/ecma262/pull/2086) | Shu-yu Guo |
     |   | 15m     | [Handle awkward rounding behavior](https://github.com/tc39/ecma402/pull/471) | Ujjwal Sharma |
+    |   | 15m     | :hourglass: [Strictness check for object's SetMutableBinding](https://github.com/tc39/ecma262/pull/2094) | Leo Balter |
```
chicoxyzzy, line 64: Does ⌛ means that the item was added after deadline? According to [https://github.com/tc39/agendas/blob/master/2020/07.md#agenda-topic-rules](agenda topic rules), any items which are not proposals looking for advancement could be added any time.

leobalter: well, I'm not fully sure but I just wanted to signal when this was added.

I'd like to address this at this meeting, but I'll understand if delegates say they didn't have enough time to review it.

ljharb: yes, and while non-proposals don't really need the hourglass, leo added it and i didn't want to remove it.

PR #829, opened 2020-07-15
30 minutes for Symbol as WeakMap key

littledan:
The discussion may be more involved


littledan:
Note that, technically, this could come earlier among the Stage 1 30-minute items (since I inserted it relatively early), but I wanted to make this item happen immediately after Records and Tuples since they're closely related.

gesa:
@littledan do me a solid and add that to scheduling constraints?

littledan:
@gesa https://github.com/tc39/agendas/pull/835

PR #853, opened 2020-09-04
fix current String.dedent proposal stage

chicoxyzzy:


ljharb (approved):
I'd changed the ordering in 216d356e; I've updated this PR to restore the proper stage 0 ordering.

PR #862, opened 2020-09-11
Late agenda item "Mechanisms to lighten the load on note-takers"

littledan:


ljharb (approved):
lateness only needs calling out for proposal advancements.

PR #876, opened 2020-10-02
Add Promise.delay and Promise.timeout

Jack-Works:
I'm not clear if I really going to add this. Leave it as a draft for now


Jack-Works:
Ok it seems like there are too many objections. I won't present this in the next meeting. 😥

PR #884, opened 2020-11-05
fix proposals sorting

chicoxyzzy:
Realms are currently on stage 2

cc @leobalter


leobalter (approved):
Thanks for fixing it!

ljharb (approved):

PR #889, opened 2020-11-09
Link to how-we-work rather than Allen's paper

littledan:
The how-we-work article is more descriptive

### DIFF
--- a/2020/11.md
+++ b/2020/11.md
@@ -12,7 +12,9 @@
 - **Location**: Remote (was Budapest)
 - **Attendee information**: https://github.com/tc39/Reflector/issues/320
 
-Allen's paper on standards committee participation for new attendees: http://wirfs-brock.com/allen/files/papers/standpats-asianplop2016.pdf
+Background:
+- Allen Wirfs-Brock's [paper on standards committee participation for new attendees](http://wirfs-brock.com/allen/files/papers/standpats-asianplop2016.pdf)
+- TC39's documentation on [How to participate in meetings](https://github.com/tc39/how-we-work/blob/master/how-to-participate-in-meetings.md)
 
 ## Agenda topic rules
 
--- a/template.md
+++ b/template.md
@@ -12,7 +12,9 @@
 - **Location**: CITY, COUNTRY
 - **Attendee information**: LINK TO REFLECTOR
 
-Allen's paper on standards committee participation for new attendees: http://wirfs-brock.com/allen/files/papers/standpats-asianplop2016.pdf
+Background:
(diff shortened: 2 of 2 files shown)


2020/11.md
```diff
@@ -12,7 +12,7 @@
 - **Location**: Remote (was Budapest)
 - **Attendee information**: https://github.com/tc39/Reflector/issues/320
 
-Allen's paper on standards committee participation for new attendees: http://wirfs-brock.com/allen/files/papers/standpats-asianplop2016.pdf
+TC39's documentation on [How to participate in meetings](https://github.com/tc39/how-we-work/blob/master/how-to-participate-in-meetings.md)
```
ljharb, line 15: I love the idea of linking to how-we-work, but I think Allen's paper is still quite helpful, and would prefer to continue linking to it.

littledan: If we link to Allen's paper, I'd like to give it context. I disagree with some things it says. For example, I don't think we should be telling new attendees "Don’t Talk Too Much". I want to encourage them to talk *more*, not less; cases of people talking too much are more likely to come from old-timers like me in my experience. The how-we-work repository gives us a place that we as a committee can edit advice to newcomers based on our values.

ljharb: What context would you suggest that we’d all agree with? Personally, i think the advice of getting the lay of the land before being outspoken is quite helpful for some newcomers (and arguably good advice for most old-timers too 😄), albeit perhaps unhelpful for those who are more introverted. I’d love a way to continue linking to the paper, while ensuring context so its advice fits a range of personality types.

rkirsling: Can't agree more with not assuming people's personalities. As someone who would have been inclined to watch how the committee operates before speaking much regardless, I believe I was less intimidated by Allen's statement than I was comforted to know that the committee values thinking before speaking.

I fully agree that the phrasing isn't ideal and I'd be happy to see us render the paper obsolete by means of how-we-work. I just think it's worth remembering that an unqualified statement of "talk more" has more potential for harm than the arguably obvious advice of "listen first".

littledan: I remain concerned with the framing of this paper. Listen first, learn background, etc are good for everyone, not just beginners. I don't agree with 100% of the tactical advice given there either. I think we should work on iterating on advice for beginners in the how-we-work repository. But, for now, if we don't have this advice reflected in the how-we-work repo, I'll leave the link for Allen's paper included in this PR.

ljharb (approved):
Thanks! this is definitely an improvement over what's in master.

PR #901, opened 2020-11-23
Add Array find right

Kingwl:


ljharb (approved):
I went ahead and added "for stage 1", since a notification of that intent is needed if that's what you're seeking. If i'm incorrect, please let me know, and I'll remove it.

PR #902, opened 2020-11-23
Add Type.isType topic

littledan:


Jack-Works:
`Type.isType(Array, arr) === Array.isArray(arr)` magically works for all built-ins across the Realm?

littledan:
Let's move discussion to https://es.discourse.group/t/strong-brand-checking-in-javascript/557

ljharb (approved):

PR #913, opened 2021-01-13
Class static init. blocks for stage 3 (tentative)

rbuckton:
Adding topic to agenda pending completion of stage 3 reviews.


ljharb (approved):
note that the slides/docs need to be present before the deadline as well.

PR #922, opened 2021-01-20
Add timeslot for Inclusion WG updates

mpcsh:
cc @davethegr8 @jasonwilliams @littledan for co-presenters, who wants in?


ljharb (approved):

ljharb:
Feel free to directly push a commit with presenter updates :-)

mpcsh:
will do @ljharb. I should've also noted that it would be best to do toward the end of the plenary so that the WG has time to prepare a good presentation

PR #936, opened 2021-02-03
Add find from right for stage 2

Kingwl:
Seems only a few works todo. I have nearly finished the spec.
I think we could push it into stage 2 after finished the spec review.


ljharb (approved):

Kingwl:
@ljharb Ahhh... Sorry for bother.

Would you mind give some suggestion? Could I try to push it into stage 2 after spec review?

Because there's seems not any others work I can do ( web compatibility?).

ljharb:
Stage 2 requires initial spec text, and all major semantics covered per https://tc39.es/process-document/.

Spec review is a stage 3 requirement.

Kingwl:
Okay. Thanks for the suggestions.

Seems It's nearly done.

PR #937, opened 2021-02-03
Add slides for array find from right

Kingwl:


ljharb (approved):
before the 10 day deadline, you're welcome to push things like this directly up as a commit.

Kingwl:
Okay. Thanks!

zloirock:
@Kingwl 2 copies of `Array.prototype.findLast` spec draft in slides instead of `Array.prototype.find`.

Kingwl:
Thanks! Fixed.

PR #942, opened 2021-02-18
Fix table for proposals nominated to stage advancement in March meeting agenda

chicoxyzzy:
Currently it looks like this
image


ljharb (approved):
thanks, seems to be from 4b5ee81975fc790466a5538ccc016ed235f97bcb

PR #953, opened 2021-03-05
Do expressions will not be asking for stage 2 yet

bakkot:


ljharb (approved):
sad :-(

Jack-Works:
what happened to the proposal? 👀

bakkot:
Just need some more time to make sure it's got the design right.

PR #978, opened 2021-04-18
Add my time constraint on 2021/04

Jack-Works:
I'd like to present on one of the following time:

- First 2hrs of any day
- If that day has Isolated Realms update, any time before it is good


ljharb (approved):

gesa:
@Jack-Works For future reference, 24 hours before meeting start is not enough time for us to be able to accommodate scheduling constraints

…I can do it today because I forgot to do some other other stuff so I happen to be at my computer on a Sunday 😬

PR #990, opened 2021-05-14
Fix markdown table

leobalter:
BTW, shouldn't the proposals be sorted by current stage?


ryzokuken (approved):

ljharb (approved):
thanks; i'll fix the sorting (and incorrect "stage" column values) in another commit

PR #993, opened 2021-05-18
Adjust timebox and add hourglass

ptomato:
I was advised that 15 minutes is the standard for a very short item and that it's better to overestimate than underestimate.


ljharb:
The hourglass only applies to stage advancement, since that's the only thing the deadline applies to.

ljharb (approved):

PR #1025, opened 2021-07-10
Module fragments slides

littledan:


littledan:
Sorry for posting these slides so late btw!

PR #1057, opened 2021-10-15
Add link, code to find local timezone start time

justingrant:
This PR:
* Clarifies the plenary meeting time zone using standard IANA names like Europe/London instead of potentially-ambiguous names like EST/BST/PST/etc
* Adds a link and a code sample to use Temporal to find out the start time of the meeting in everyone's local time zone and localized date/time format


bakkot:
Let's just put both BST and Europe/London and land this as-is, and we can worry about more complete and automatic solutions later.

bakkot:
It's worth having at least some way of getting the right time, this method is cute and harmless, and it helps familiarize delegates with Temporal's API.

Also, multiple delegates have been confused about the start time for prior meetings, especially around DST boundaries. Time zones are hard!

PR #1058, opened 2021-10-15
Adds slides for Partial Application

rbuckton:


rbuckton:
Running up against agenda deadline, so I will merge these directly.

rbuckton:
I've had a proposal rejected solely on the grounds I'd created the branch but neglected to hit submit on the PR until after the deadline. Not quite the same since the visibility is lower on a branch than a PR, but I didn't want to take the chance.

ljharb:
Definitely not the same at all, but understood.

rbuckton:
Also, the agenda submission deadlines are the following:

image

The only mention of only having a pull request is in regards to proposals not looking to advance. For all other bullets it says "added", which seems to convey the meaning that they must be present in the actual agenda. If just having the pull request is acceptable to avoid this restriction, I suggest we change the wording under "Agenda topic rules" to reflect this.

ljharb:
Sure, done in d2ef8097

PR #1079, opened 2021-11-03
Fix Temporal snippet in the 2021-12 agenda

nicolo-ribaudo:


ptomato (approved):
Should probably also delete the comment that says "DELETE THIS COMMENT", it's only supposed to be part of the agenda template :smile:

nicolo-ribaudo:
Done!

PR #1094, opened 2021-12-09
Reduce timebox for RegExp set notation topic + add slides

mathiasbynens:


ljharb:
@mathiasbynens no need for a label; GitHub has a native draft feature. Mark it as ready for review when it is so :-)

PR #1108, opened 2022-01-14
Add class brand check topic to next meeting agenda

hax:
Add class brand check to agenda


ljharb (approved):
squeaking it in just before the deadline :-p

hax:
> squeaking it in just before the deadline :-p

Ah, it seems we do not have much topics this time, so I added it in last minute :-p

PR #1116, opened 2022-01-19
Rename string split proposal

lucacasonato:


ljharb (approved):
typically proposals should be named after the problem they solve, rather than the specific solution proposed - especially prior to stage 2

PR #1133, opened 2022-03-18
2022-03: Add JSC items

js-choi:
• Bikeshedding pipe operator’s topic reference (at Stage 2, no advance)
• Bikeshedding bind-this syntax (at Stage 1, no advance)
• Function.prototype.once for Stage 1
• Holistic discussion of TC39 dataflow proposals


rkirsling:
I was like "wait, what do these have to do with JavaScriptCore?" for like 30 seconds before I remembered your initials... 😅

ljharb (approved):
fixed the sorting

PR #1141, opened 2022-03-25
Not requiring stage advancement for pattern matching on this meeting

Jack-Works:
we have no spec for pattern matching, therefore not be able to advance to stage 2. change it to an update


ljharb (changes requested):
That's not accurate (we have initial spec text, which is all stage 2 requires), and by removing the request from the agenda, we'd *definitely* be unable to advance. Let's please leave this here.

PR #1153, opened 2022-03-29
Withdrawing Destructuring Privates

jridgewell:
We require a bit more holistic design with the eventual private declarations and object support: https://github.com/tc39/proposal-destructuring-private/issues/9


littledan:
I'm looking forward to working together with @jridgewell on this broader concept once I'm back from my break.

PR #1154, opened 2022-05-16
adding symbols as weakmap keys for stage 3

acutmore:
Need to discuss with champion group who is going to present.

Stage 2 spec review has not finished. Added on the assumption that this will be complete before plenary.


ljharb (approved):
landing now, feel free to directly push an update when you now who's presenting

PR #1168, opened 2022-05-27
Adds entry for RegExp Atomic Operators

rbuckton:
Taking another run at Atomic Operators


jridgewell:
Slide 7 has a typo in the shorthand section, I'm assuming they're meant to be `(?>…)`

rbuckton:
Ah good catch. Fixed, thanks!

PR #1169, opened 2022-05-27
Adds slides for Grouped and Auto-Accessors

rbuckton:


bathos:
Random drive by: I hadn’t looked at this proposal before and I was really impressed by the clarity of the slides & the proposed design. I would find this very valuable personally — hope it advances!

PR #1177, opened 2022-05-31
Add DurationFormat stage 3 update

ryzokuken:
Apologies for being behind the deadline.


ljharb (approved):
The deadline does not apply to things that aren't asking for advancement or normative changes.

PR #1180, opened 2022-06-01
Fix decorators slides link

pzuraq:
cc @ljharb I think it got messed up in the rebase


ljharb (approved):
yep, wasn't the rebase, it was my vim copy-paste (－‸ლ) thanks!

PR #1184, opened 2022-06-05
Fix rendering of findLast/findLastIndex

robpalme:
I am assuming DanielR wants to present given he added this.


ljharb (approved):
good catch!

PR #1195, opened 2022-07-06
Add topic about WebAssembly Polyfills

sffc:
CC @robpalme 

This is a presentation I've been wanting to make for a while. I would prefer to do it in the September meeting, but I can put together some slides for the July meeting if it is better.


sffc:
Thinking about it more, I will be able to make a much better presentation if I have more lead time. I cannot commit to making a polished presentation in time for the July plenary. I will consider adding it to the September or November meeting.

PR #1206, opened 2022-07-09
2022-07: Object.pick (explainer only)

js-choi:


hemanth:
Closing this on behave of a repo in the transfer org which is still cooking.

PR #1240, opened 2022-09-02
Add set methods agenda item

bakkot:
Didn't link the proposal repo 'cause it's stale and I don't have write access.


ljharb:
@bakkot since you now have write access, can we link to the proposal repo?

bakkot:
I need a few hours to update the repo. Let's merge this for now and I'll follow up adding the link when I'm content with it. The slides themselves are the important part; nothing I'm going to be discussing depends on the repo.

ljharb (approved):

PR #1241, opened 2022-09-02
String.dedent discussion

jridgewell:
To discuss https://github.com/tc39/proposal-string-dedent/pull/60


bakkot:
String.dedent is at stage 2, not 3, I think?

jridgewell:
Whoops

ljharb (approved):
fixed ordering

PR #1261, opened 2022-11-16
Add module expressions&declarations

nicolo-ribaudo:
Part of the discussion will probably apply to both proposals, so I would like to have them one after the other. I was considering putting them under a single agenda item, but since:
- they are at different stages
- one is a status update, the other one is asking for advancement

I ended up using two topics with a schedule constraint that I would like to have them together. It's not exactly how schedule constraint are meant to be used: is it ok?

Also I moved the Module/ModuleSource proposal to the correct position in the table, since it's not at Stage 2 yet (cc @kriskowal / @caridy)


ljharb (approved):

kriskowal:
Thank you, @nicolo-ribaudo and @ljharb for cleaning up after my stumbles. Genuinely appreciated and I’ll try not to rely upon your efforts in the future!

PR #1274, opened 2022-11-23
2022.11: Move schedule constraints to the correct sub-section

nicolo-ribaudo:
As far as I understand, all constraints should go either in the "Normal constraints" or in the "Late-breaking schedule constraints" sections.

It's important that they are properly sorted, because as the comment in the agenda says "Conflicting constraints honored on a first-come, first served basis". This PR sorts them in the same order as they were committed.


ljharb (approved):
thanks

PR #1277, opened 2022-11-23
Reduce Array Grouping Timebox

jridgewell:


ljharb:
30 minutes seemed ambitious; what are you hoping to discuss in only 15 minutes?

jridgewell:
The discussion doesn't need to be blocking, it can be discussed and come back later in the meeting with a decision.

PR #1294, opened 2022-12-17
Update link for the submission deadline

bertyhell:
to match 10 am est 30th jan 2023


ljharb (approved):
thanks, missed that one

PR #1295, opened 2023-01-06
Add proposed dates for the rest of the year to the top-level README

tabatkins:


ljharb (changes requested):
This is in an autogenerated section of the readme (`npm run build`); it also would create a secondary source of truth that needs updating, so i'm not sure it's the best place to put it.

ljharb:
Although maybe we should move it to the readme _above_ the autogenerated section, and remove it from the agendas themselves (and replace them with a link to the readme section)?

tabatkins:
Ah, had no idea it was auto-genned, but now that comment in the source makes sense. I've moved the table to its own details above the auto-gen section for now; we can think about exposing this information to the autogen later.

ljharb:
Sounds great! Can you also remove the table from the template and the January agenda, and make both link to the new table in the readme?

michaelficarra:
Also, the January meeting is still upcoming!

tabatkins:
Jan meeting is upcoming, but confirmed and with its own page already, so it doesn't go in the "proposed" table. ^_^

jridgewell:
> Jan meeting is upcoming, but confirmed and with its own page already, so it doesn't go in the "proposed" table. ^_^

I would make this a "future meetings" section instead of a "**tentatively planned** future meetings", with a "confirmed" field on the table

tabatkins:
I added a "Confirmed" column, but since I waited so long, the Jan meeting is no longer upcoming at all. ^_^

README.md
```diff
@@ -2,6 +2,18 @@ agendas
 =======
 
 TC39 meeting agendas
+
+<details>
+<summary>Future Meetings</summary>
+
+| Confirmed | Future Dates             | Location            | Host                 |
+|-----------|--------------------------|---------------------|----------------------|
+|           | 2023-03-21 to 2023-03-23 | Seattle, US         | F5                   |
```
ljharb, line 11: i think this one's pretty confirmed

tabatkins:
Cool, put a checkmark in its row now

jridgewell (approved):

PR #1297, opened 2023-01-12
Request TC39 adoption of W3C Guidelines

pipobscure:
I ask the plenary to make the [W3C Design Guidelines](https://w3ctag.github.io/design-principles/) advisory precedent which should be taken into account by all proposals to enhance consistency across the web-plattform.

Given this is already a standard document by W3C we ask the chairs to host the discussion. A conclusion would be consensus and clarity on how the plenary believes these design guidelines should impact TC39 work and whether and how they should influence us.

### DIFF
--- a/2023/01.md
+++ b/2023/01.md
@@ -65,6 +65,7 @@ Supporting materials includes slides, a link to the proposal repository, a link 
 
     | ✓ | timebox | topic | presenter |
     |:-:|:-------:|-------|-----------|
+    |   |15-30    | Adopt [W3C Guidelines](https://w3ctag.github.io/design-principles/) as advisory precedent for TC39 | Chairs |
 
 1. Overflow from previous meeting
 


ctcpip:
to be clear, it's NOT a standards document, its status is "Group Note", and the distinction is important

see: https://www.w3.org/standards/types#NOTE

> A W3C Group Note is a document produced by a W3C Working Group, a W3C Interest Group, the Advisory Board (AB), or the W3C Technical Architecture Group (TAG). A W3C Group Note is a [W3C Technical Report](https://www.w3.org/TR/).

> A Group Note is to provide a stable reference for a document that is not intended to be a formal standard. These notes have not received formal review and are not endorsed W3C.

> These notes MUST NOT be cited as W3C standards and may or may not become W3C Statements.

> Software MAY implement these reports at their own risk. Implementation is neither discouraged nor encouraged but can contribute to proposals for further action on a specification.

> There are no patent protection covering the implementations of the Group Note.

ljharb:
It'd still be something we could discuss in plenary, but either way I'm not in favor of subordinating ourselves to another standards body on something that's not strictly necessary (like CLDR, eg)

bakkot:
FWIW I doubt that formally adopting these guidelines would do us much good - it would just be another thing to fight over - but I've personally found them thoughtful and occasionally clarifying in the past when thinking about proposals myself. So while I wouldn't relish trying to get the committee to adopt them in any formal sense, it might be worth having a 5 or 10 minute item that is less "let us spend some time fighting about whether to bind ourselves to this document" and more "here is a document I like and a brief summary of important points; you may wish to use it to guide your own thinking".

PR #1310, opened 2023-01-20
Add JSON.parse source text access update

gibson042:


ljharb (approved):

michaelficarra:
@gibson042 @ljharb This was added in the wrong position. Until we have CI enforcement, please be careful to check that the agenda rule for proposal-related item ordering is followed.

zloirock:
(I hope that it will be added as new methods since for many cases required sequential semantics.)

PR #1319, opened 2023-01-31
Add topic for decorator/export ordering for January 2023 meeting.

DanielRosenwasser:
Given the discussion in https://github.com/tc39/proposal-decorators/issues/69, this topic is something we really need to discuss for the January meeting, since it will inform what we ship in TypeScript 5.0 RC.


ljharb (approved):
heads up to the @tc39/chairs

PR #1322, opened 2023-02-12
Schedule constraint for March

guybedford:
Thought I'd get this one in now. Would be great if timings can work out for modules agenda items, but as permits of course.


ljharb (approved):

nicolo-ribaudo:
I suggest explicitly writing in the constraint that you want to be present for modules, so that chairs don't have to guess or find this PR.

PR #1324, opened 2023-02-19
add agenda item for Float16Array for stages 2 and 3

bakkot:


ljharb (approved):

phoddie:
I'm unsure if advancing a proposal two stages in a single meeting is consistent with the stage process. I acknowledge that it may not be expressly disallowed by the process text, but it seems to side-step the deliberate pace intended by the stage process.

leobalter:
IIRC, this already happened in the past, but I also remember the message was that double advancement could be rejected if a delegate express the desire of keeping the one-step pacing for the proposal to reach better maturity.

At the same time, @phoddie it seems like from your comment this might not be ready for implementation. I'm supporting you here if that's the case.

ljharb:
it's indeed happened many times before and is often acceptable when semantics are straightforward and noncontroversial, but as @leobalter indicates, it's also totally fine to withhold consensus for two stages at once if there's a reason to go slower.

phoddie:
> it's indeed happened many times before

phoddie:
> it's indeed happened many times before

Thanks, @ljharb. While I've not attended everything TC39 meeting in the last 5 years, I've been at plenty. And while my memory is far from perfect, I don't recall this happening. I do recall it being proposed a couple times. Would you mind citing an example or two?

>  it's also totally fine to withhold consensus for two stages at once if there's a reason to go slower

Understood and appreciated. Still, this assumes that a delegate who might object is fully engaged with TC39. The reality is that not everyone has time to track every proposal at every meeting. Advancing proposals by no more than one stage at a meeting gives more delegates an opportunity to provide feedback. That's fundamental to the stage process. And given how controversial proposing a step backwards in the stage process is, a measured paced is all-the-more important.

> t seems like from your comment this might not be ready for implementation. I'm supporting you here if that's the case.

@leobalter, thank you. I'm not taking a position here on the proposal itself. At first look, it appears straightforward.

ljharb:
@phoddie 
 - https://github.com/tc39/proposals/commit/309b3ac2b30d19bc9a84cca6c7c90e137f6f03b5 Duplicate named capture groups went from 0 to 2, conditional on spec review
 - https://github.com/tc39/proposals/commit/11a23823a906bff22c2bca0167519d055c57fead Destructuring private fields went directly to stage 2
 - https://github.com/tc39/proposals/commit/ea97e3d87bbd14cddb06a9b20714b034369ccc59 `Object.hasOwn` went directly to stage 2
 - https://github.com/tc39/proposals/commit/d34899051b78bc3f68b5fb15c975fac1fdf2abf6 hashbang grammar went directly to stage 2
 - https://github.com/tc39/proposals/commit/812c9f370ec891363301ce3feeb3f94097f5915d ArrayBuffer transfer went directly from inactive to stage 2
 - https://github.com/tc39/proposals/commit/a6696379bcec50cea8b3a32496dcaa4d00baf636 Promise.prototype.finally went directly to stage 2
 - https://github.com/tc39/proposals/commit/8a333a1cb9a0d301401e7608cf776c7d8b7a3ef4 went directly to stage **3**
 
That's all I could find with some trivial code searches of the proposals repo.

PR #1358, opened 2023-04-10
Discuss Array.prototype.group

littledan:
Volunteers needed!


ljharb:
I believe @jridgewell was waiting on compat data before adding this one.

littledan:
@ljharb Is there any sort of effort ongoing to get compat data? I think we should at least have a status update, even if it's not conclusive yet.

ljharb:
@littledan yes, that's what's being waited on.

michaelficarra:
This was added out of order. Fixed in https://github.com/tc39/agendas/commit/c947ab82b3305349a61eafbfe6ec4f4d19661193.

michaelficarra:
Also, @littledan please make sure you have a presenter listed before the meeting starts. If we don't have a presenter for each topic, I won't be able to approve the agenda.

PR #1366, opened 2023-04-24
🚨 lint fixes, change config to allow inline html

ctcpip:


ChristianUlbrich (approved):
Looks good to me. :) Cold not find any other disabling of `MD033`in the repo. Seems like the _default_ is to [not allow anything at all](https://github.com/DavidAnson/markdownlint/blob/main/doc/md033.md), we might configure this rule, to allow only a sub set of tags (`sub`, `sup`, ...), instead of _full-blown_ HTML. But I am fine with _full-blown_ HTML. :)

PR #1368, opened 2023-05-04
2023.05: Add "Module Harmony: interaction semantics"

nicolo-ribaudo:
I'm not 100% sure about where to put this topic in the agenda. It's about proposals, but not a single one so I wouldn't know which stage to pick in that section.


ljharb (approved):
This seems fine for now; chairs can move if needed.

PR #1380, opened 2023-05-10
Add discussion item for decorator initializer order

pzuraq:
This adds an item for discussing decorator initializer order. This has recently been brought up by users of decorators as an issue with consistency that makes it difficult to translate some existing decorators, and to maintain logically consistent behavior between initializers and setters.

This item is NOT asking for any normative changes to the spec, is it is past the deadline. It is moreso to bring the issue to the committee's attention and discuss possible resolutions, or if a resolution is necessary. If it would be more appropriate to add this to the longer/open-ended discussions section I would be happy to do that instead.


ljharb (approved):
note that you can still ask for normative changes past the deadline, it's just that consensus _can_ be withheld solely on the basis of not meeting the deadline.

PR #1391, opened 2023-05-15
Add slides for decorator initializer order discussion

pzuraq:
Adds some slides to help summarize/clarify the ongoing discussion


nicolo-ribaudo:
Could you link to slides that we can scroll through, rather than the "live" version, so that people can read them in advance?

pzuraq:
updated!

PR #1392, opened 2023-05-16
🔧 markdownlint cli

ctcpip:


package.json
```diff
@@ -1,11 +1,14 @@
...
+    "build": "./scripts/generate.js",
+    "//": "lint command ignores old files",
+    "lint": "markdownlint-cli2 \"**/*.md\" \"#node_modules\" \"#201[0-9]/*.md\" \"#202[0-2]/*.md\" \"#2023/01.md\""
   },
   "devDependencies": {
     "dedent": "^0.7.0",
     "glob": "^7.2.3",
-    "markdown-magic": "^0.1.25"
+    "markdown-magic": "^0.1.25",
+    "markdownlint-cli2": "^0.7.1"
```
ljharb, line 12: btw why cli2 instead of cli? i use cli on another repo

ljharb (approved):

package.json
ctcpip: probably nothing wrong with `cli` but `cli2` is from the same maintainer as `markdownlint`

PR #1401, opened 2023-06-20
Add Temporal agenda item

ptomato:


michaelficarra (approved):
we should just add this to the template at this point 😅

PR #1423, opened 2023-07-05
📝 necessary -> supporting

ctcpip:
`necessary materials` is ambiguous and undefined.  this change normalizes the text to `supporting materials` which is the correct term used throughout and is defined


ljharb (approved):
I don't think it's ambiguous, since the definition of what materials are needed exists, but the change seems fine

ctcpip:
well, it was ambiguous enough that multiple people reviewing the agenda were uncertain whether `necessary` and `supporting` materials referred to the same thing 🤷

ljharb:
Fair enough; i don't think that feedback had come up before.

michaelficarra (approved):

PR #1428, opened 2023-07-10
Justin Grant's schedule constraints (late-breaking)

justingrant:


justingrant:
Never mind, I just saw proposed schedule and it looks fine. Closing.

ctcpip:
@justingrant It was I who had the constraint for those items for the afternoon, but it's more of a preference than an absolute must.  I may be able to shuffle things around to place those in morning slots.  If the existing schedule looks ok to you, then great, but just throwing this out there in case it's problematic.

justingrant:
@ctcpip - Nah, it's OK. If I have to stay up until 4:00AM anyways for the Temporal presentation, then I can make it until 4:30 for the proposal that I'm presenting. Will be a fun party.

ctcpip:
@justingrant wait, what TZ are you in?

PR #1449, opened 2023-09-17
Add slides for operators, literals

littledan:
Literals must come after operators


ljharb:
usually the constraint covers that; the ordering in the table is supposed to strictly be by stage decreasing and timebox increasing, but w/e

michaelficarra:
Agreed with @ljharb. We should stick to the proposal ordering rules, even when there's interdependencies.

PR #1451, opened 2023-09-18
Add ECMA-402 PR 788

sffc:
CC @gibson042 

Added it with an hourglass.


ctcpip:
should we move this up to be with the other 402 items?

ljharb (approved):
it's sorted correctly (by timebox, ascending) but it could be presented in any order you like ofc

ctcpip:
ah right

PR #1470, opened 2023-10-21
Add call for JSX

Jack-Works:
https://app.element.io/#/room/#jsx:matrix.org

### DIFF
--- a/2023/11.md
+++ b/2023/11.md
@@ -106,6 +106,7 @@ When applicable, use these emoji as a prefix to the agenda item topic.
 
     | ✓ | timebox | topic | presenter |
     |:-:|:-------:|-------|-----------|
+    |   | 2m  | Recruiting people interested in JSX to join the [matrix room](https://matrix.to/#/#jsx:matrix.org) | Jack Works |
     |   | 10m | add stage names back to process document ([PR](https://github.com/tc39/process-document/pull/36)) | Michael Ficarra |
 
 1. Proposals


NullVoxPopuli:
Can jsx continue to be as application specific thing?
Too many people disagree with it for it to be a platform feature.

It also eliminates this from being a possibility: https://wycats.github.io/polaris-sketchwork/content-tag/second-phase-compiler.html

Jack-Works:
I do not have a proposal for now, I am just trying to make a place for interested TC39 members to discuss it. If it becomes a proposal, of course it will be not React-specific. If you're using a preprocessor to handle JSX (as I understand from the link you posted), it should continue to work _even_ if it is incompatible with the final proposal. It just like how Svelte build a compiler that use complete different semantics than what we have in JavaScript.

NullVoxPopuli:
well, I was more hoping we pursue `content-tag` as a proposal, as it enables not only JSX, but all other transpilation as well.

like, in userland, you could have the JS syntax support all of:
```js
import { jsx } from 'jsx-processor';

let count = 1;

<jsx>
  <div>
    {count}
  </div>
</jsx>
```

so, if we reserve the `<tag>` syntax for something _more generic_ than JSX, than we can inherently support all transforms, and not perpetuate react/jsx favoritism.

for example, we could have:

```js
import { gql } from 'gql-processor';

<gql>
    {
      user(id: 5) {
        firstName
        lastName
      }
    }
</gql>
```

In both of these cases, `gql` and `jsx` are functions that receive the string, and is _not_ interpolatable (that would be the application-specific post-processing).

So, for example:
```js
function jsx(jsxTemplate) {
  return convertJsxTemplateToReactCreateComponent(jsxTemplate);
}
```
or
```js
function gql(gqlTemplate) {
  return convertGqlTemplateToQuery(gqlTemplate);
}
```

This would allow way more long-term flexibility, and would be much more supportive of a broader spectrum of JS ecosystems.

----------------

AST-wise, this could be:
```js
import { jsx } from 'jsx-processor';

let count = 1;

<jsx>
  <div>
    {count}
  </div>
</jsx>
```
becomes:
```js
📂 File
  📂 Program
    📁 ImportDeclaration
    📁 VariableDeclaration
    📂 ExpressionStatement
      📂 expression: ContentTag
        📂 id: Identifier
            name: 'jsx'
        📂 content: StringLiteral
            value: '<div>...\n</div>'
```

2023/11.md
```diff
@@ -120,6 +120,7 @@ When applicable, use these emoji as a prefix to the agenda item topic.
 
     | ✓ | timebox | topic | presenter |
     |:-:|:-------:|-------|-----------|
+    |   | 2m      | Recruiting people interested in JSX to join the [matrix room](https://matrix.to/#/#jsx:matrix.org) | Jack Works |
```
ctcpip, line 123: if the timebox is correct (2 minutes), this belongs in the above section: `Short (<=30m) Timeboxed Discussions`

Jack-Works:
@NullVoxPopuli if it does not support interpolation, it will be useless for JSX's use case, and I also want to note that we're not proposing a proposal, just having people have room to discuss. you can open a new issue in the JSX proposal repo in the future, but now it's too early

ljharb (approved):

NullVoxPopuli:
@Jack-Works if you need interpolation, why not use backticks? :thinking:

ljharb:
That would only produce a string

PR #1476, opened 2023-11-02
Add note to put schedule constraints in the relevant section

nicolo-ribaudo:
If we have the "normal constraints" section we should use it


ctcpip (commented):
forgot to hit approve before merging... approved FTR

PR #1478, opened 2023-11-10
link attendee information, fix date in temporal snippet

bakkot:
We should make a checklist for copying over the templates, given how many of these little details need to be gotten right.

### DIFF
--- a/2023/11.md
+++ b/2023/11.md
@@ -24,7 +24,7 @@ gantt
 For meeting times in your timezone, visit [Temporal docs](https://tc39.es/proposal-temporal/docs/) and run the code below in the devtools console.
 
 ```js
-Temporal.ZonedDateTime.from('2023-11-17T10:00[America/Los_Angeles]')
+Temporal.ZonedDateTime.from('2023-11-27T10:00[America/Los_Angeles]')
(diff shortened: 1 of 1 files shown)


2023/11.md
```diff
@@ -24,7 +24,7 @@ gantt
 For meeting times in your timezone, visit [Temporal docs](https://tc39.es/proposal-temporal/docs/) and run the code below in the devtools console.
 
 ```js
-Temporal.ZonedDateTime.from('2023-11-17T10:00[America/Los_Angeles]')
```
michaelficarra, line 27: As written, this gives you the advancement deadline. I think that's fine.

bakkot: That would be fine if it said "for _the advancement deadline_ in your timezone", but it claims to be the meeting time.

ctcpip:
> We should make a checklist for copying over the templates, given how many of these little details need to be gotten right.

the agenda and the reflector issue are separate activities.  when there's a `.race()`, and the agenda appears first, the reflector link isn't available.

bakkot:
Sure, but it's not just that one; there's a bunch of dates and etc which need to be updated.

ctcpip:
clipgenda

ctcpip (approved):

PR #1488, opened 2023-11-20
Update 11.md

pzuraq:
Adds an agenda item for discussing a normative change to the decorators proposal (After deadline I realize, but wanted to bring it to the attention of the committee regardless)


ljharb (approved):
fixed ordering and added hourglass

PR #1539, opened 2024-02-01
Update 2024-02 ShadowRealms to update only

leobalter:
ShadowRealms will be presented as a proposal update only, no longer seeking stage advancement at this meeting.

cc @caridy @dminor @syg @ptomato


ljharb (approved):
fixed ordering

PR #1544, opened 2024-02-04
don't hide future meeting list behind <details>

michaelficarra:


ljharb:
This seems like it was done pretty intentionally in #1295.

tabatkins:
It was intentional, yes, but only because the other meeting sections were all also in details. I don't have a problem with someone changing it if they think it reads better to be always visible.

That said, the heading structure in this PR is busted; it puts all the *past* meetings under the "Future Meetings" heading. You'll need to add a `## Past Meetings ##` header too.

michaelficarra:
Headings fixed.

ljharb (approved):

PR #1554, opened 2024-03-20
🔧 add npm test

ctcpip:


.github/workflows/lint.yml
```diff
@@ -12,5 +12,4 @@ jobs:
       - uses: actions/checkout@v3
       - uses: ljharb/actions/node/install@main
         name: 'nvm install lts/* && npm install'
-      - run: npm run lint
-
+      - run: npm test
```
ljharb, line 15: i would still expect the lint workflow to run the lint script

ctcpip: if I rename the workflow, does that satisfy?

ljharb: i mean i guess - what's the benefit tho?

(i'm on board with making `npm test` run all the things locally, but you almost never want that in a single CI job)

ctcpip: ok

ljharb (approved):

PR #1561, opened 2024-03-27
🔥 remove image

ctcpip:


ljharb:
why remove it?

ctcpip:
one might argue we shouldn't store corporate campus maps in the agenda repo

ljharb:
google earth exists, and if a campus map that's been public for 7 years is a security issue that suggests bigger problems :-) did someone at paypal request this?

littledan:
This seems like a privacy-driven removal suggestion. @SaminaHusain is responsible for such matters legally for Ecma, as our GDPR point-person (not sure of the official term). Maybe she will have thoughts here. IMO this would be a good patch to land.

robpalme (approved):

PR #1610, opened 2024-06-01
Adding TCQ reloaded status update

ChristianUlbrich:
Just _a little_ bit late to the game. I blame soccer (for @robpalme football)! Let's hope, no one notices!


robpalme (approved):
This is not proposal advancement, so it's fine to land this after the final whistle.

PR #1617, opened 2024-06-04
Add `await using`

rbuckton:
Adding overflow topic as we did not have time to come back to the deterministic collapse of Await in `await using` topic in the last plenary session.


ljharb (approved):

ctcpip:
technically not overflow as we concluded the topic.  either way, this will be high enough on the agenda to not get bumped, so it doesn't matter too much

ctcpip (changes requested):
I'll move the item

ctcpip (approved):

rbuckton:
We didn't conclude the topic. IIRC, I had requested a timebox extension to discuss Nicolo's feedback but we didn't get to it.

rbuckton:
Per the first paragraph at https://github.com/tc39/notes/blob/main/meetings/2024-04/april-11.md#continuation-extractors, we couldn't continue the "Deterministic collapse of Await" portion of that discussion because Nicolo was not present.

rbuckton:
If it needs to be a new topic, that's fine, but I don't have independent slides for just that topic as I'd planned to just continue from where I left off in the slides from the previous session.

ctcpip:
tl;dr:  this is practically pedantic in nature, and I am now regretting saying anything and wasting everyone's time on it 🫠

oh no, who is `Nickoli`? 😅   -- that was during the extractors continuation, which was indeed requested.  there was not a request for ex. res. mgmt.  I don't think anything needs to be a 'new topic' -- this is just for schedule prioritization (which constraints take precedence over anyway).

ctcpip:
https://github.com/tc39/notes/pull/327

rbuckton:
My bad for not having corrected the transcription there, I overlooked the spelling. 

My only concern about moving it is just how it's interpreted. As an overflow topic, it's obvious that I'm just continuing the topic from last session. As a new topic I feel it would be better to have a more specific name and slides. I can put up a different PR later today.

ctcpip:
Oh, ok.  I didn't realize that continuity aspect was at play here.  That's totally fine, I will bring it back to where you had it.

ctcpip:
for your consideration: https://github.com/tc39/agendas/pull/1618

PR #1637, opened 2024-07-01
Scrub

littledan:


ptomato:
Last meeting, I volunteered to facilitate one of the scrub sessions. I might be unavailable for some of the days of the July meeting, but if it happens on a day where I'm present, I'd be happy to present this one.

ljharb (approved):

nicolo-ribaudo:
If it happens in a day when Philip is not available, I can also facilitate.

PR #1648, opened 2024-07-18
add Intl.DurationFormat: Display negative sign on leading numeric-style zeroes

ben-allen:
Previous versions of the spec failed to account for the possibility that negative durations could start with zero-valued hours or minutes values.


ljharb (approved):
fixed the sorting

PR #1664, opened 2024-07-26
peter klecha presenter constraints

peetklecha:


ljharb (approved):

ctcpip:
hmm.. it's unlikely that we will get to this topic earlier than the 31st

peetklecha:
is the 1st not an option?

ctcpip:
We are unlikely to go into day 4, due to amount of content.

PR #1676, opened 2024-09-03
Add Observables to agenda

domfarolino:


2024/10.md
```diff
@@ -126,6 +126,7 @@ When applicable, use these emoji as a prefix to the agenda item topic.
     |:-------:|-------|-----------|
     | 60m | Explicit compile hints | Marja Hölttä (Google) |
     | 90m | JSSugar/JS0 | Shu-yu Guo |
+    | 45m | WHATWG Observables | Dominic Farolino |
```
ctcpip, line 129:
```suggestion
    | 45m | WHATWG Observables | Dominic Farolino |
    | 60m | Explicit compile hints | Marja Hölttä (Google) |
    | 90m | JSSugar/JS0 | Shu-yu Guo |
```

ctcpip: please fix ordering, per rules (shorter timebox items come first)

ctcpip (dismissed):

2024/10.md
ljharb: done.

ljharb (approved):

ctcpip (approved):

domfarolino:
I meant to type this up earlier but I got wrapped up in a meeting:

I have a few questions before committing to adding this to the agenda. I see the f2f is 3 days long, but is there any guarantee on which day I would be able to present Observables? (I ask because October 10th is the one day in that range that does not work for me). Furthermore, I will be presenting the proposal remotely from East Coast, USA. This will be challenging timezone-wise, so I'm wondering if there is a concrete time slot that I can know I'll be presenting in?

ljharb:
You can add "schedule constraints" where you request the chairs slot your item in specific times or days.

PR #1677, opened 2024-09-07
Add @domfarolino's presentation constraints

domfarolino:


2024/10.md
```diff
@@ -126,7 +126,7 @@ When applicable, use these emoji as a prefix to the agenda item topic.
 
     | timebox | topic | presenter |
     |:-------:|-------|-----------|
-    | 45m | WHATWG Observables | Dominic Farolino |
+    | 45m | 🔒 WHATWG Observables | Dominic Farolino |
```
domfarolino, line 129: Daniel Minor's entry (also subject to scheduling constraints) does not include this emoji, but I think if I'm reading the emoji table correctly, this is the right formatting?

ljharb (approved):

2024/10.md
ljharb: you're right, both should

PR #1680, opened 2024-09-10
Add "Immutable ArrayBuffers" and "Eval Twin problem"

erights:


ljharb:
@erights any links?

erights:
Thanks. The first link is good. I also updated the second with a good interim link for now, but hope to replace with something better soon.

Done.

ljharb (approved):

PR #1706, opened 2024-10-02
Experiment: Add total duration to the agenda

nicolo-ribaudo:
image

Lets see if this works -- I'm not sure about how GitHub caches images.

Server code (using temporal and iterator helpers 😄): https://dash.deno.com/playground/tc39-agenda-time

```js
const pattern = new URLPattern({ pathname: "/:year/:month{/}?" });

Deno.serve(async (req: Request) => {
    try {
        const { year, month } = pattern.exec(req.url).pathname.groups;

        const res = await fetch(`https://raw.githubusercontent.com/tc39/agendas/refs/heads/main/${year}/${month}.md`);
        const body = await res.text();

        let outputText;

        if (res.status === 404) {
            outputText = "Err 404"
        } else if (!res.ok) {
            return new Response("Error while fetching the agenda:\n\n" + body, {
                status: res.status,
                headers: { "content-type": "text/plain" }
            });
        } else {
            const minutes = body.matchAll(/\b(\d+)m(?=\s*[,|)])(?!\) Timeboxed Discussions)/g)
                .map(m => parseInt(m[1], 10))
                .reduce((sum, m) => sum + m, 0);

            const duration = Temporal.Duration.from({ minutes }).round({ largestUnit: "hours" });

            outputText = `${duration.hours}h ${duration.minutes}m`;
        }

        return new Response(`
            <svg version="1.1" width="80" height="20" xmlns="http://www.w3.org/2000/svg">
                <style>
                    @media (prefers-color-scheme: dark) {
                        text {
                            fill: white;
                        }
                    }
                </style>
                <text x="0" y="18" font-family="sans-serif">${outputText}</text>
            </svg>
        `, {
            headers: { "content-type": "image/svg+xml", "cache-control": "no-cache, max-age=0" }
        })
    } catch {
        return new Response("Error", { status: 500 });
    }
});
```

If this works, then I can add it to the template.

### DIFF
--- a/2024/10.md
+++ b/2024/10.md
@@ -9,6 +9,7 @@
   - 10:00 to 17:00 JST (Asia/Tokyo) on 10 October 2024
 - **Location**: Tokyo, Japan
 - **Attendee information**: https://github.com/tc39/Reflector/issues/537
+- **Total duration of scheduled discussions**: ![dynamically computed total of timeboxes on the agenda](https://tc39-agenda-time.deno.dev/2024/10/)
 
 ```mermaid
 gantt
--- a/2024/12.md
+++ b/2024/12.md
@@ -10,6 +10,7 @@
   - 10:00 to 15:00 MST (America/Denver) on 5 December 2024
 - **Location**: Albuquerque, New Mexico
 - **Attendee information**: https://github.com/tc39/Reflector/issues/545
+- **Total duration of scheduled discussions**: ![dynamically computed total of timeboxes on the agenda](https://tc39-agenda-time.deno.dev/2024/12/)
 
 ```mermaid
 gantt
--- a/2025/02.md
+++ b/2025/02.md
@@ -9,6 +9,7 @@
   - 10:00 to 16:00 PST (America/Los_Angeles) on 20 February 2025
 - **Location**: Seattle, WA
 - **Attendee information**: https://github.com/tc39/Reflector/issues/547
+- **Total duration of scheduled discussions**: ![dynamically computed total of timeboxes on the agenda](https://tc39-agenda-time.deno.dev/2025/02/)
 
 ```mermaid
 gantt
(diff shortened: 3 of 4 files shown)


nicolo-ribaudo:
Should we extract only the durations in the tables, or (which is what this PR is doing) also those in the bullet points?

ljharb:
Github caches images very aggressively, and I'd think we want to avoid adding more external dependencies.

nicolo-ribaudo:
Wdyt about an action that updates the time whenever there is a new commit on main?

ljharb:
That seems like it'd create a ton of noise commits :-/

bakkot:
Are we worried about the number of commits in this repo? I certainly would prioritize having the total time readily available over minimizing the number commits, personally.

ljharb:
In general yes, git commit history is worth keeping clean on every repo.

I'm sure we can come up with a workaround - a github pages page that calculates it, for example, or a CI check that blocks additions that exceed the total time - that doesn't require spamming commits to the default branch.

nicolo-ribaudo:
It seems like the `cache-control` header is working properly to prevent GitHub from caching the image — the preview of this PR already updated the total time due to https://github.com/tc39/agendas/pull/1707

ryzokuken:
It doesn't work as well for dark mode 
image

nicolo-ribaudo:
@ryzokuken Fixed

ryzokuken (approved):
Looks great now

bakkot:
For alt text, something like "dynamically computed total of timeboxes on the agenda"?

nicolo-ribaudo:
Alt text outdated. Also I figured out how to change the URL to something nicer.

michaelficarra:
Note that this will possibly show the wrong duration when looking at older versions of the agenda.

ljharb:
how so? the year and month are part of the URL

michaelficarra:
@ljharb Not other agenda documents. An older commit.

nicolo-ribaudo:
Yeah this will always fetch the duration from main. That part of the agenda is mostly meant for us so it's probably fine, but I can add an explicit mention of this.

nicolo-ribaudo:
PR updated to also add the duration to the agenda of the upcoming meeting.

ryzokuken:
Can we merge this now?

ljharb (approved):

PR #1671, opened 2024-08-12
Update 10.md

marjakh:

PR #1707, opened 2024-10-02
Update 10.md

marjakh:
Moving my talk to the 30 min list. Since I added it to the agenda earlier than the other 30 min talks, adding it in the second relative position (if that matters - not sure).

( https://github.com/tc39/agendas/pull/1671/files was where it was originally added)


ljharb (approved):
LGTM

PR #1729, opened 2024-10-30
📝 embrace unicode in the year 2024

ctcpip:


mhofman:
I would not have tried to use a unicode char in the branch name ! Wondering how many tools that breaks.

ctcpip:
let this be a lesson to them!

PR #1789, opened 2025-02-06
add links to spec PRs for going-for-stage-4 proposals

bakkot:
"Proposals looking to advance to stage 4 must link to a pull request into [the spec](https://github.com/tc39/ecma262), since the [process](https://tc39.github.io/process-document/) requires one."

(In both cases you can find the spec PR by following a link from the linked issues, but I have always understood the requirement to link to supporting materials to mean that you must _directly_ link to them.)


ljharb (approved):
i don't think "directly" is part of the requirement, but it's fine to do so ofc

michaelficarra:
FWIW I understood this requirement to mean "directly" as well. @tc39/chairs I'm sure could amend the wording with notice to TG1.

PR #1813, opened 2025-03-02
Fix the date

hax:
The start date should be April 14?

### DIFF
--- a/2025/04.md
+++ b/2025/04.md
@@ -17,17 +17,17 @@ gantt
   1 month before the meeting: milestone, 2025-03-14 10:00-0500, 1m
   Proposal advancement deadline: milestone, 2025-04-04 10:00-0500, 1m
   Schedule constraint deadline: milestone, 2025-04-11 10:00-0500, 1m
-  Day 1 : day1, 2025-04-15 10:00+0000, 7h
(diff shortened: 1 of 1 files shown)


2025/04.md
```diff
@@ -27,7 +27,7 @@ gantt
 For meeting times in your timezone, visit [Temporal docs](https://tc39.es/proposal-temporal/docs/) and run the code below in the devtools console.
 
 ```js
-Temporal.ZonedDateTime.from('2025-04-15T10:00[America/New_York]')
+Temporal.ZonedDateTime.from('2025-04-14T10:00[America/New_York]')
```
ljharb, line 30: i'm confused, the first day of the meeting is the 15th, not the 14th - it seems correct as-is? maybe the "1 month before" date needs to be fixed to the 15th though

bakkot: @ljharb The "Dates and times" at the top of this doc lists the 14th. And since this is a remote plenary, and the 14th is a Monday, that seems like it's right to me?

ljharb: ahhh ok, the dates are wrong in the mermaid diagram also. i'll take care of it as part of this PR.

ljharb (approved):

PR #1816, opened 2025-03-26
Update 04.md for three proposals

erights:
Added three advancement placeholder entries that need links, after more work is done on the content.

Before the advancement deadline, for each I will either fill in links to good enough content, or I will remove from this agenda.


ljharb (approved):
0 to 2.7, wow, looking forward to the links

PR #1836, opened 2025-04-07
Remove Allen's standards participation guide

littledan:
I don't want us to tell new delegates, "Don’t Talk Too Much", which comes up very prominently and early in this document. Instead, we should encourage new delegates to talk! We can reference this document from some places, as it has interesting ideas, but only if we can frame it with that context.

### DIFF
--- a/template.md
+++ b/template.md
@@ -32,10 +32,7 @@ Temporal.ZonedDateTime.from('2021-10-25T10:00[Europe/London]')
   .toLocaleString()
 ```
 
-Background:
-
-- Allen Wirfs-Brock's [paper on standards committee participation for new attendees](http://wirfs-brock.com/allen/files/papers/standpats-asianplop2016.pdf)
-- TC39's documentation on [How to participate in meetings](https://github.com/tc39/how-we-work/blob/HEAD/how-to-participate-in-meetings.md)
+TC39's documentation on [How to participate in meetings](https://github.com/tc39/how-we-work/blob/HEAD/how-to-participate-in-meetings.md)
 
 ## Agenda topic rules
 


ljharb (changes requested):
There's still a lot of useful information in that paper even if you don't agree with the general advice to listen a lot before talking when entering any new arena - i think if you want to encourage new delegates to talk, it might be better if you created a third resource to add to this list, perhaps as the first item?

littledan:
We created how-we-work for this.

ljharb:
Sure, but that doesn't mean this paper isn't still worth recommending. Where in how-we-work is there a counterpoint to the "listen before talking" advice?

phoddie:
As a new delegate, I found Allen's guide to be incredibly helpful, even encouraging. I never took it as an admonition not to participate, but rather to understand the norms of the conversation before jumping in. That advice is all the more valuable now because TC39 is so much bigger.

linusg:
If someone in committee told me "don't talk too much" in those words I would forward that to the CoC committee for being disrespectful and unprofessional. Why are we recommending a paper that does it?

Raising concerns as a newcomer is intimidating enough as is, if someone does all of the paperwork needed to attend plenary in a non-observing capacity they have automatically "earned" their right to speak. TCQ and the chairs ensure it happens in an orderly manner.

michaelficarra:
As a big PDF, we don't get to pick and choose parts or rewrite parts as our standards for participation change over time. For that reason alone, I would prefer to get rid of it and replace it with a living document that we can all stand behind.

bakkot:
I'd prefer not to remove it without a good replacement, but until such a thing exists we might add a note about this not being fully endorsed by the current committee? Possibly something like:

> Allen Wirfs-Brock's paper on standards committee participation for new attendees, though some of the advice should not be taken to apply to the modern TC39 - in particular, _contra_ this paper, active participation from new delegates is encouraged; you can pick up a lot of norms from reading [our documentation](https://github.com/tc39/how-we-work/blob/HEAD/how-to-participate-in-meetings.md) and from participating rather than sitting on the sidelines when starting out.

ctcpip:
I agree with just about every take in this discussion.  Namely:

- the guide is useful / helpful
- the guide is imperfect and can give people the wrong impression
- the guide is immutable and in a less accessible format

It's also not permissively licensed, so we could not do something like convert it to markdown and make changes, PRs with reviews, etc.

It's worth noting that in our [Welcome Letter](https://github.com/tc39/Reflector/wiki/Onboarding-message) to new members, we direct people to Jory's [Revised Patterns for Participation in Standards Committees](https://medium.com/@jorydotcom/revised-patterns-for-participation-in-standards-committees-dae82d93954e)

That guide is also subject to some of the issues of the former (e.g. we can't copy and revise it, it's hosted on a paywalled website, etc.)

We could reach out to Allen and/or Jory if we wanted to go down the path of adopting and updating, or just create/augment our own version.

I do think there is information in there that is not well-covered by `how-we-work` content, and I am apprehensive about removing the reference when we don't have a comprehensive alternative to direct folks to.

ryzokuken:
I had originally thought this PR removes all references to AWB's paper which I could understand the pushback to (although I'm in favor of it given all the concerns listed by others above). But now I realize (somehow noticed it for the first time ever) that we're removing the link from every single agenda item which makes a lot of sense. Why do we include it on every single agenda doc ever? I'm pretty sure nearly nobody ever follows the link and reads the doc from there and we are already emailing it to everyone as part of their intro email anyway.

ljharb:
@ryzokuken because that long predates "how we work", so there wasn't any other place to put it.

I suspect it would be uncontroversial to *move* it to how we work, though.

ryzokuken:
I'd strongly support moving it from agenda to how-we-work or even to the agendas README as opposed to every individual agenda, in the spirit of keeping the individual agenda docs smaller and relevant.

PR #1840, opened 2025-04-10
Add schedule constraints

sffc:
The GitHub UI was going to allow me to commit directly to the `main` branch. Should I do that now?


ljharb (approved):
you could, but since it's 3 commits, i'll rebase it down and land it :-)

PR #1856, opened 2025-05-06
Immutable ArrayBuffer for stage 3

erights:
(I may have committed something to main accidentally. Sorry, but this would overwrite that anyway.


ljharb (approved):
i'd pushed a fix commit to add the stage and timebox; i just rebased this PR to use your chosen timebox and update presenters

PR #1864, opened 2025-05-13
Add AsyncContext web integration brainstorm session

andreubotella:


ljharb (approved):

nicolo-ribaudo:
Chairs, should this go in "Longer or open-ended discussions"? The champions are not actually proposing _something_, it's more going to be a "the web integration needs to be simplified, how do committee members value various part of it and what can we give up?"

ljharb:
It certainly could, but there's no requirement for it to.

PR #1891, opened 2025-05-28
Propose Friday breakout session for JS numerics (decimal, measure/amo…

jessealama:
…unt, etc.)


jessealama:
> Can we have 90 minutes? :)

I bumped the requested timebox to 90 minutes.

robpalme (approved):

PR #1910, opened 2025-07-18
add Freezing the Array Iterator item

bakkot:


ljharb (approved):

zloirock:
This is a good suggestion, but it needs to be seriously tested to see if it will break the web. I'm not sure about some of my code -)

> Maybe also freeze ArrayIteratorPrototype and the above two functions to avoid making a  communication channel if people care about that?

I remember some proposals whose polyfills would require patching `ArrayIteratorPrototype`. For example, https://github.com/leebyron/ecmascript-reverse-iterable. I'm pretty sure that some other future conceptions will require `ArrayIteratorPrototype` extending.

bakkot:
I agree with both the point about needing to carefully assess web compat and the risk of making future proposals harder to polyfill. Ability to polyfill is a tradeoff like everything else - it's definitely a good but it may be worth giving it up in some circumstances. Remains to be seen what the tradeoffs here are. Personally I don't really care about the communications channel thing but if that's the only way to get this done at all I'd take that tradeoff even if it does make some possible future proposals impossible to polyfill.

PR #1912, opened 2025-07-18
Add import hook and new global

kriskowal:
We’re tentatively preparing to *at most* seek advancement of Module Import Hooks and new Globals to Stage 1, but likely more of a status update for the existing Stage 1 compartment proposal


kriskowal:
All set. Thanks.

PR #1915, opened 2025-07-18
2025.07: Add keep-trailing-zeros slides and add Amount for Stage 2

eemeli:
I also noticed that the Import Buffer item appears to have been added in the wrong place in the agenda, but didn't touch it.


nicolo-ribaudo:
Technically it was `Object.propertyCount` that was added in the wrong place :P

PR #1922, opened 2025-07-21
🔧 optimize CI runs

ctcpip:
eliminate duplicate runs on PRs

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,6 +1,11 @@
 name: 'Lint Markdown'
 
-on: [pull_request, push]
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+  workflow_dispatch:
 
 permissions:
   contents: read


michaelficarra (approved):

ljharb:
I'd left it this way intentionally, since it's not a given that a pushed branch will immediately be turned into a PR. The runs are also parallelized, so this won't make full runs any faster, it'll just save github some money.

michaelficarra:
Do people look at actions status for commits that aren't part of some PR? I'd think this would be a waste somewhere between 99% and 100% of the time.

ljharb:
I do, but either way it's only a waste for github ¯\\\_(ツ)_/¯

ctcpip:
> it's not a given that a pushed branch will immediately be turned into a PR

do we really care about pushed non-default branches running CI right away?  either they will get a PR, or they will get merged to main without a PR and if that happens, it's likely nobody was looking at the CI result anyway

it's not a big deal either way, but I have my preference 😄

michaelficarra:
I'm okay with improving things for GitHub.

UlisesGascon (approved):

PR #1925, opened 2025-07-24
Please schedule Import Hook and New Global together

kriskowal:


naugtur:
note that slides link already exists in a separate PR

PR #2084, opened 2026-05-06
Add CRA primer and Q&A

gesa:


michaelficarra (approved):
Love it. Hopefully there will be slides that we can share with our colleagues.

PR #2090, opened 2026-05-08
2026.05: Add links for Stable Formatting + Add Default Behaviours for some Intl APIs for Stage 1

eemeli:


eemeli:
I included a second agenda item in this proposal, as it ends up on the immediately following line and is in any case a spin-off from Stable Formatting. Adding both at the same time avoids a merge conflict.

jessealama (approved):

PR #2106, opened 2026-05-15
fix: module-global is Stage 1

naugtur:
I noticed there was a mistake - the proposal was advanced to Stage 1 before.


ljharb (approved):
fixed ordering

PR #2114, opened 2026-06-23
s/SafeResolve/SafePromiseResolve/

mgaudet:
Forgot what I had titled the AO. 🤦


linusg:
I'd just drop that part entirely, the issue of proposals (and thus agenda items and meeting notes) changing name because they are named after a function or AO subject to bikeshedding comes up again and again. I think "Thenable Curtailment" is sufficient.

mgaudet:
Heh -- I just wrote a slide about renaming :P I am OK with keeping the name tho. I'll update this branch in a second.

PR #2122, opened 2026-07-10
add private declarations item

bakkot:


acutmore (approved):
Side note: destructuring? For these and also private fields?

bakkot:
Yes [I also want that](https://github.com/tc39/proposal-destructuring-private) (though I wouldn't say it's "and also" private fields, these are just private fields).

PR #2127, opened 2026-07-14
add GitHub Action to check proposal agenda item ordering

michaelficarra:
In its current configuration, this will create inline annotations for any discovered violations using GitHub's [workflow commands](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands). In the future when we're more confident in the lack of false positives, we can pass `--strict` to fail CI for any violations.

Disclaimer: I used AI in the process of creating this PR.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,4 +1,4 @@
-name: 'Lint Markdown'
+name: 'Lint and Test'
 
 on:
   push:
@@ -17,4 +17,4 @@ jobs:
       - uses: actions/checkout@v3
       - uses: ljharb/actions/node/install@main
         name: 'nvm install lts/* && npm install'
-      - run: npm run lint
+      - run: npm test
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -14,7 +14,7 @@ jobs:
   lint:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v7
       - uses: ljharb/actions/node/install@main
         name: 'nvm install lts/* && npm install'
       - run: npm test
--- a/package.json
+++ b/package.json
@@ -6,7 +6,8 @@
     "check-proposals-order": "node ./scripts/check-proposals-order.cli.mjs",
     "lint": "markdownlint-cli2",
     "lint:fix": "markdownlint-cli2 --fix",
-    "test": "npm run lint && npm run test:unit",
+    "pretest": "npm run lint",
+    "test": "npm run test:unit",
     "test:unit": "node --test 'scripts/*.test.mjs'"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "scripts": {
     "a2s": "node ./scripts/agenda-to-schedule.mjs",
     "build": "./scripts/generate.js",
-    "check-proposals-order": "node ./scripts/check-proposals-order.mjs",
+    "check-proposals-order": "node ./scripts/check-proposals-order.cli.mjs",
     "lint": "markdownlint-cli2",
     "lint:fix": "markdownlint-cli2 --fix",
     "test": "npm run lint && npm run test:unit",
--- a/package.json
+++ b/package.json
@@ -3,9 +3,11 @@
   "scripts": {
     "a2s": "node ./scripts/agenda-to-schedule.mjs",
     "build": "./scripts/generate.js",
+    "check-proposals-order": "node ./scripts/check-proposals-order.mjs",
     "lint": "markdownlint-cli2",
     "lint:fix": "markdownlint-cli2 --fix",
-    "test": "npm run lint"
+    "test": "npm run lint && npm run test:unit",
+    "test:unit": "node --test 'scripts/*.test.mjs'"
   },
   "devDependencies": {
     "dedent": "^1.7.0",
--- a/package.json
+++ b/package.json
@@ -6,7 +6,6 @@
     "check-proposals-order": "node ./scripts/check-proposals-order.cli.mjs",
     "lint": "markdownlint-cli2",
     "lint:fix": "markdownlint-cli2 --fix",
-    "pretest": "npm run lint",
     "test": "npm run test:unit",
     "test:unit": "node --test 'scripts/*.test.mjs'"
   },
--- a/scripts/check-proposals-order.cli.mjs
+++ b/scripts/check-proposals-order.cli.mjs
@@ -51,7 +51,10 @@ const SEVERITY = {
 const RUN_FAILURE_KINDS = new Set(['missing-table']);
 
 function git(...args) {
-  return execFileSync('git', args, { encoding: 'utf8', stdio: ['ignore', 'pipe', 'pipe'] });
+  return execFileSync('git', args, {
+    encoding: 'utf8',
+    stdio: ['ignore', 'pipe', 'pipe'],
+  });
 }
 
 function tryGitShow(commit, file) {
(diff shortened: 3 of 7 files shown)


scripts/check-proposals-order.mjs
```diff
@@ -0,0 +1,385 @@
...
+// the first link URL in the topic cell, used to re-identify a row whose text
+// was edited (e.g. a slides link was added)
+function firstTopicUrl(cells) {
+  const match = /\]\(\s*<?([^)>\s]+)/.exec(cells[2] ?? '');
+  return match == null ? null : match[1];
+}
+
+// finds every proposals table: a `| stage | timebox | topic | presenter |`
+// header (the only 4-column table in an agenda) followed by a delimiter row
+export function parseProposalsTables(contents) {
```
acutmore, line 58: Would it be worth asserting that we find a table, to avoid the structure changing and the match silently failing?

michaelficarra: Yes, and done.

package.json
```diff
@@ -3,9 +3,11 @@
   "scripts": {
     "a2s": "node ./scripts/agenda-to-schedule.mjs",
     "build": "./scripts/generate.js",
+    "check-proposals-order": "node ./scripts/check-proposals-order.mjs",
     "lint": "markdownlint-cli2",
     "lint:fix": "markdownlint-cli2 --fix",
-    "test": "npm run lint"
+    "test": "npm run lint && npm run test:unit",
```
ljharb:
```suggestion
    "pretest": "npm run lint",
    "test": "npm run test:unit",
```

.github/workflows/lint.yml
```diff
@@ -17,4 +17,4 @@ jobs:
       - uses: actions/checkout@v3
```
ljharb:
```suggestion
      - uses: actions/checkout@v7
```
might as well, while we're in here

scripts/check-proposals-order.mjs
```diff
@@ -0,0 +1,417 @@
...
+  if (match == null) return null;
+  if (match[3] != null) return Number(match[3]) * 60;
+  return (match[1] == null ? 0 : Number(match[1]) * 60) + Number(match[2]);
+}
+
+// the first link URL in the topic cell, used to re-identify a row whose text
+// was edited (e.g. a slides link was added)
+function firstTopicUrl(cells) {
+  const match = /\]\(\s*<?([^)>\s]+)/.exec(cells[2] ?? '');
+  return match == null ? null : match[1];
```
ljharb:
```suggestion
  return match?.[1];
```

scripts/check-proposals-order.mjs
```diff
@@ -0,0 +1,417 @@
...
+    }
+  }
+  // a failure to run takes precedence over ordering violations
+  if (runFailed) process.exit(2);
+  if (options.strict && errorCount > 0) process.exit(1);
+}
+
+if (process.argv[1] != null && import.meta.url === pathToFileURL(process.argv[1]).href) {
+  main();
+}
```
ljharb: this is pretty awkward; can the parts that need testing live in a different file than the one expected to be ran on the CLI?

package.json
michaelficarra: Why? And if I did this, why not inline test:unit into here?

ctcpip:
we already pull in `marked` as a dependency. consider using that to cut down on the bespoke parsing

scripts/check-proposals-order.mjs
michaelficarra: Split.

michaelficarra:
@ctcpip No, that would actually be a lot more complicated, and there's also no preserved position information (used for error reporting) in marked's parsed output. Also, it mangles the "raw" values in its token stream, so accurately reconstructing positions is virtually impossible.

package.json
ljharb: why, because `&&` can cause issues on windows; and why not, because then you couldn't run test:unit reliably without running the lint script.

scripts/check-proposals-order.cli.mjs
```diff
@@ -0,0 +1,159 @@
...
+    console.error(`OK: ${file}`);
+  } else if (annotations.some(annotation => RUN_FAILURE_KINDS.has(annotation.kind))) {
+    console.error(`Could not check ${file}: no proposals table found.`);
+  } else {
+    console.error(`${annotations.length} ordering problem(s) introduced in ${file}`);
+  }
+}
+// a failure to run takes precedence over ordering violations
+if (runFailed) process.exit(2);
+if (options.strict && errorCount > 0) process.exit(1);
```
ljharb:
```suggestion
if (runFailed) process.exitCode = 2;
else if (options.strict && errorCount > 0) process.exitCode = 1;
```
you're not generally supposed to call exit() anymore, but obv in this case it doesn't really matter

ljharb (approved):

package.json
michaelficarra: I'm afraid to ask, but how to people on Windows do `&&`?

scripts/check-proposals-order.cli.mjs
michaelficarra: Done.

scripts/check-proposals-order.cli.mjs
```diff
@@ -0,0 +1,200 @@
...
+function escapeAnnotationProperty(text) {
+  return escapeAnnotationData(text).replace(/:/g, '%3A').replace(/,/g, '%2C');
+}
+
+// an IIFE so the early bailouts below can set process.exitCode and `return`;
+// setting the code rather than calling process.exit lets buffered annotation
+// output flush before the process ends (it can be truncated otherwise in CI)
+(() => {
+  const options = { base: null, head: null, strict: false, files: [] };
+  const argv = process.argv.slice(2);
```
bakkot: It doesn't actually matter, but, `util.parseArgs` has existed since node 18.3 and is better than this.

ljharb: sooo much cleaner with it, good suggestion

scripts/check-proposals-order.cli.mjs
```diff
@@ -0,0 +1,199 @@
...
+(() => {
+  let options, positionals;
+  try {
+    ({ values: options, positionals } = parseArgs({
+      options: {
+        base: { type: 'string' },
+        head: { type: 'string' },
+        strict: { type: 'boolean', default: false },
+      },
+      allowPositionals: true,
```
bakkot, line 90:
```suggestion
      allowPositionals: true,
      allowNegative: true,
```
just on general principle (I don't expect anyone will actually do `--no-strict`).

scripts/check-proposals-order.mjs
```diff
@@ -0,0 +1,280 @@
...
+// was edited (e.g. a slides link was added)
+function firstTopicUrl(cells) {
+  const match = /\]\(\s*<?([^)>\s]+)/.exec(cells[2] ?? '');
+  return match?.[1];
+}
+
+// finds every proposals table: a `| stage | timebox | topic | presenter |`
+// header (the only 4-column table in an agenda) followed by a delimiter row
+export function parseProposalsTables(contents) {
+  const lines = contents.replace(/\r\n/g, '\n').split('\n');
```
gibson042: Nit:
```suggestion
  const lines = contents.split(/\r?\n/g);
```

gibson042 (approved):

michaelficarra:
@tc39/chairs Is this something I can merge now that it's gone through review and has some approvals?

package.json
```diff
@@ -3,9 +3,12 @@
   "scripts": {
     "a2s": "node ./scripts/agenda-to-schedule.mjs",
     "build": "./scripts/generate.js",
+    "check-proposals-order": "node ./scripts/check-proposals-order.cli.mjs",
     "lint": "markdownlint-cli2",
     "lint:fix": "markdownlint-cli2 --fix",
-    "test": "npm run lint"
+    "pretest": "npm run lint",
```
ctcpip: not a big fan of `pretest` because it requires knowledge of how/when it runs. I'd rather be explicit by bundling it in `test`, though I did see the discussion about `&&` ...

.github/workflows/lint.yml
```diff
@@ -14,7 +14,7 @@ jobs:
   lint:
```
ctcpip:
```suggestion
  lint-and-test:
```

ctcpip: maybe rename the file too?  obv these are very minor things and I am not blocking on any of this

package.json
michaelficarra: I don't know how I'm supposed to reconcile these conflicting requests.

michaelficarra: I made it so that `test` doesn't run `lint` anymore. They're separate scripts. The CI action runs both in series.

.github/workflows/lint.yml
michaelficarra: Done.

package.json
ctcpip: I don't know, I think the likelihood of `&&` being problematic on windows is low enough that this is fine actually

ctcpip: meaning:

`"test": "npm run lint && npm run test:unit",`

👍

ljharb: ¯\_(ツ)_/¯ it's fine, but it's idiomatic to use pre and post scripts and confine each script to a single command.

package.json
ljharb: "pretest" doesn't "require knowledge", it's how every npm package works since 2009.

michaelficarra:
Alright it seems like we're no longer talking about the substance of this PR. Unless anyone protests, I'm going to merge this tomorrow (or whenever I next remember it exists, whichever is later). You can iterate on your npm scripts in another PR.

